PR #67, opened 2020-08-24
Add capsule-user-group CLI flag

### LINKED ISSUE #66: Support for custom user group
# Describe the feature

Hi guys, thank's a lot for a great tool for managing multi-tenant k8s :) 

We are now trying to adopt it in our environment and found that right now it's impossible to use custom UserInfo.Group in webhooks so they apply only for specific users, as it's hardcoded as `capsulev1alpha1.GroupVersion.Group`. As we already have a dedicated AD group for k8s users, it will be good if we can use it with Capsule too, without a need to provision one more group and move users into it.

I will submit a PR with this small patch.

Thanks,
Max

MaxFedotov:
This PR closes #66 and adds additional configuration parameter `capsule-user-group` which will be used in webhooks to filter out users for whom these webhooks apply.

It's completely backward-compatible, as default  `capsule-user-group` is set to `capsulev1alpha1.GroupVersion.Group`


prometherion:
This PR looks very good, thanks @MaxFedotov : may I ask you to implement a simple e2e or simply an unit test to cover this case?

prometherion (approved):

MaxFedotov:
@prometherion 
Need some advice about organizing e2e tests. Right now it's impossible to add custom arg for manager (e.g. set `capsule-user-group`) in order test different flags, they all get overridden by contents of  `default\manager_auth_proxy_patch.yaml` when running `make deploy`. 

So we need an idea about how to run `make deploy` with different capsule flags

MaxFedotov:
Or for example, we can just override spec for Deployment with custom args for a test and then rollback it back to defaults :)

prometherion:
I think we could easily add an additional test case specific for this feature, editing the Capsule deployment at test runtime using the `JustBeforeEach` and `JustAfterEach` Ginkgo's helpers.

At test case tear up, we could edit the Deployment `args` in order to add the requested parameters, to turn them down at tear down.

The Capsule user group is used mainly for Webhooks since we're intercepting these requests to block some features, so what we could do is to duplicate some of the actual test specs or, rather, edit the actual ones forcing this.

If you're not comfortable about what I wrote and asked (I needed some time to understand properly Ginkgo and not yet sure I understood it tho :sweat_smile:) we can merge this and fix the missing tests afterward: the great thing is this stunning feature implementation, so many kudos again!

prometherion (commented):
Ok, I have to change my idea regarding this due to a blocker.

Actually, we're using the User Group to determine if a request is originated by a Capsule user, and that's ok. Although, that's not enough since we're using the hard-coded value [here](https://github.com/clastix/capsule/blob/0f935d53b7acb17379e749a2d64e2f631adb6779/config/tenants/namespace-provisioner.yaml#L17-L18) at manifest installation (as you pointed correctly with the `make deploy`).

I think that the whole folder containing the Cluster Role and related Bindings should not be handled as static manifests but rather controlled at runtime by the Operator, also to ensure the desired state is matched despite changes or deletions.

I'm going to merge this once fixed this really good catch, hopefully upon the next week since I'm actually on vacation 🏖️

prometherion (changes requested):
Issue #69 must be fixed first.

MaxFedotov:
@prometherion 
Yeah, you are right. I can try to implement this, it's not a big deal. I don't think that there is a need for a dedicated controller for this, just a simple function, which is called before `mgr.Start(ctrl.SetupSignalHandler())` and checks `namespace:provisioner` CRB and paches subject if it needed. What do you think about such approach?

 Btw `namespace:deleter` can remain unchanged, as there is no CRB, it is used only to create namespaced RB

MaxFedotov:
@prometherion 
So maybe implement it the same way, as TLS\CA controllers? Where empty secrets are created with kustomize and later updated by manager?

prometherion:
I would implementing something similar but different: we need the scaffolded secret since we have to get it mounted to server Dynamic Admissions requests (aka Webhooks).

`controllerutils.CreateOrUpdate` is pretty sure the func we need, along with Predicates as you could have saw from the Secret Controller in order to get informed only about specific resources.

I guess we don't need to get these CR/B entities managed by Kustomize or Helm so managing at runtime is neat.

MaxFedotov:
Ok, will implement it today. Should I add it as a commit to this branch or create new one?

prometherion:
I would keep these separated, so I would open a PR addressing especially #69

MaxFedotov:
> I would implementing something similar but different: we need the scaffolded secret since we have to get it mounted to server Dynamic Admissions requests (aka Webhooks).
> 
> `controllerutils.CreateOrUpdate` is pretty sure the func we need, along with Predicates as you could have saw from the Secret Controller in order to get informed only about specific resources.
> 
> I guess we don't need to get these CR/B entities managed by Kustomize or Helm so managing at runtime is neat.

So you are talking about separate controller for CR/B. But i'm curious, how it would work, as these type of resources may not exist when manager starts, so there will be no events to trigger Reconcile. So either we should emit them (some type of generic event on startup, which will be picked up by controller), or just use a simple function in order to create them if they not exists during startup

MaxFedotov:
@prometherion Can you please take a look on my second commit? I've decided to add it to this branch, so I don't need to rewrite code and add support for `capsule-user-group` by merging two branches, because they seem to relate to the same problem :)

Thanks,
Max

pkg/rbac/rbac.go
```diff
@@ -0,0 +1,90 @@
...
+
+type RBACManager struct {
+	ClientConfig *rest.Config
+	CapsuleGroup string
+	Log          logr.Logger
+}
+
+func (r RBACManager) SetupCapsuleRoles() error {
+	// Create client as mgr.Client won't be availiable for us until we call mgr.Start
+	k8sClient, err := client.New(r.ClientConfig, client.Options{})
```
prometherion, line 40: why importing `k8sClient` here using the `*rest.Config`?
You can use the `client.Client` interface provided by `mgr.GetClient()` in order to use also the local cache.

pkg/rbac/rbac.go
```diff
@@ -0,0 +1,90 @@
...
+	r.Log.Info("setting up ClusterRoleBindings", "ClusterRoleBinding", provisionerRoleName)
+	if err := k8sClient.Get(context.TODO(), types.NamespacedName{Name: provisionerRoleName}, clusterRoleBinding); err != nil {
+		if errors.IsNotFound(err) {
+			if err = k8sClient.Create(context.TODO(), provisionerClusterRoleBinding); err != nil {
+				return err
+			}
+		}
+	} else {
+		// RoleRef is immutable, so we need to delete and recreate ClusterRoleBinding if it changed
+		if !equality.Semantic.DeepDerivative(provisionerClusterRoleBinding.RoleRef, clusterRoleBinding.RoleRef) {
```
prometherion, line 76: nice catch here using `equality` 💪🏻

main.go
```diff
@@ -105,6 +107,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (rbac.RBACManager{
```
prometherion, line 110: I think we need here a reconcile function, although the `SetupCapsuleRoles` is pretty good since we have to force the first reconciliation.

I would go implementing the `reconcile.Reconciler` interface, forcing an initial reconciliation at setup time.

pkg/rbac/const.go
```diff
@@ -0,0 +1,67 @@
...
+
+package rbac
+
+import (
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	provisionerRoleName = "capsule-namespace:provisioner"
```
prometherion, line 25: I guess we need to export the `deleterRoleName` constants since we need it here:
https://github.com/clastix/capsule/blob/0f935d53b7acb17379e749a2d64e2f631adb6779/controllers/tenant_controller.go#L510-L521

Or rather, we could move that piece of code under this package: what do you think?

prometherion (changes requested):
These changes look promising, just keep in mind we have to update also the [README](https://github.com/clastix/capsule/blame/master/README.md#L127-L128) regarding the output upon installation since these Cluster Roles are going to be managed by the Capsule controller.

prometherion:
According to the latest changes, this can fix #69

pkg/rbac/rbac.go
MaxFedotov: mgr.GetClient() won't be available until manager starts, as it relies on caches. It was my first try to use it - but I won't be able to GET resources due to error

main.go
MaxFedotov: That was my first intention too - create a custom controller to manage CR\B. But then I thought it would be too complex for a such simple task, as every controller consumes compute resources. These CR\B are quite static and I don't see any use case when someone will update them manually, so this controller won't be capturing any events. The only way for them to change - is to modify Deployment Pod Spec in order to add some manager flags, and this will cause pod recreating and CR\B will be updated if needed.

pkg/rbac/const.go
MaxFedotov: I will make the exportable :)

pkg/rbac/rbac.go
prometherion: I guess we can overcome this with the following [PR](https://github.com/wgsateam/capsule/pull/1)

e2e/utils_test.go
```diff
@@ -20,18 +20,21 @@ package e2e
...
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/clastix/capsule/api/v1alpha1"
 )
 
 const (
-	defaultTimeoutInterval = 10 * time.Second
+	defaultTimeoutInterval = 20 * time.Second
```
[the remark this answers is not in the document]
MaxFedotov: I had to do it because we are recreating a pod in order to apply `capsule-user-group` config option and on my mac it can require more than 10 seconds :)

pkg/rbac/rbac.go
MaxFedotov: thx :)

pkg/rbac/rbac.go
MaxFedotov: Great catch, thanks! Will merge it right now

main.go
MaxFedotov: From my point of view, human can ruin everything - for example modify namespace labels with a wrong label, remove services and so on. It's impossible to protect from all human errors, and I really don't think that dealing with operators mistakes should be something that Capsule is in charge of :) 
But that's up to you to deside.

bsctl:
> What I'm more concerned is the change at runtime performed by an human, you know: shit happens, and dealing with it in an automated way rather than restarting the controller would be perfect, although the computational cost.

If I well understood the question, the suggestion is to make it enough robust and find a tradeoff with the computational cost, of course. Please, consider that one of first use cases for this project is to build a public CaaS. A robust CR/B control is necessary, imho: 

> shit happens

MaxFedotov:
@prometherion @bsctl 
Ok, agreed. Maybe you can merge this PR and create an additional issue for CR\B controller?

prometherion (commented):
LGTM, I would like to have [this](https://github.com/wgsateam/capsule/pull/2) implemented too that essentially put the RBAC manager as provisioned and controller, too.

prometherion:
Tests are red but testing locally and seem ok, so we can ignore that since it seems bogus.

```
creating a Namespace over-quota
  should fail
  /home/prometherion/Documents/capsule/e2e/overquota_namespace_test.go:53
STEP: creating three Namespaces
 
• [SLOW TEST:13.206 seconds]
creating a Namespace over-quota
/home/prometherion/Documents/capsule/e2e/overquota_namespace_test.go:32  
  should fail
  /home/prometherion/Documents/capsule/e2e/overquota_namespace_test.go:53
------------------------------ 
exceeding Tenant resource quota
  should block new Pods if limit is reached
  /home/prometherion/Documents/capsule/e2e/resource_quota_exceeded_test.go:136 
STEP: creating the Namespaces  
STEP: retrieving the Resource Quota  
STEP: ensuring the status has been blocked with actual usage 
STEP: creating an exceeded Pod 
STEP: retrieving the Resource Quota  
STEP: ensuring the status has been blocked with actual usage 
STEP: creating an exceeded Pod 
 
• [SLOW TEST:5.138 seconds] 
exceeding Tenant resource quota
/home/prometherion/Documents/capsule/e2e/resource_quota_exceeded_test.go:39 
  should block new Pods if limit is reached
  /home/prometherion/Documents/capsule/e2e/resource_quota_exceeded_test.go:136 
------------------------------ 
creating a Namespace trying to select a third Tenant
  should fail
  /home/prometherion/Documents/capsule/e2e/selecting_non_owned_tenant_test.go:53  
STEP: assigning to the Namespace the Capsule Tenant label 
•
------------------------------ 
when Tenant owner interacts with the webhooks 
  should disallow deletions 
  /home/prometherion/Documents/capsule/e2e/owner_webhooks_test.go:103 
STEP: blocking Capsule Limit ranges  
STEP: blocking Capsule Network Policy
STEP: blocking blocking Capsule Resource Quota
•
------------------------------ 
when Tenant owner interacts with the webhooks 
  should allow listing
  /home/prometherion/Documents/capsule/e2e/owner_webhooks_test.go:147 
STEP: Limit Range resources 
STEP: Network Policy resources 
STEP: Resource Quota resources 
•
------------------------------ 
when Tenant owner interacts with the webhooks 
  should allow all actions to Tenant owner Network Policy resources
  /home/prometherion/Documents/capsule/e2e/owner_webhooks_test.go:182 
STEP: creating  
STEP: updating  
STEP: deleting
•
------------------------------
creating a Namespace as Tenant owner with custom --capsule-group
  should fail [60/305]
  /home/prometherion/Documents/capsule/e2e/custom_capsule_group_test.go:54  
 
• [SLOW TEST:22.158 seconds]
creating a Namespace as Tenant owner with custom --capsule-group
/home/prometherion/Documents/capsule/e2e/custom_capsule_group_test.go:32 
  should fail
  /home/prometherion/Documents/capsule/e2e/custom_capsule_group_test.go:54  
------------------------------ 
creating a Namespace as Tenant owner with custom --capsule-group
  should succeed and be available in Tenant namespaces list  
  /home/prometherion/Documents/capsule/e2e/custom_capsule_group_test.go:61  
 
• [SLOW TEST:23.184 seconds]
creating a Namespace as Tenant owner with custom --capsule-group
/home/prometherion/Documents/capsule/e2e/custom_capsule_group_test.go:32 
  should succeed and be available in Tenant namespaces list  
  /home/prometherion/Documents/capsule/e2e/custom_capsule_group_test.go:61  
------------------------------ 
Namespace creation with no Tenant assigned 
  should fail
  /home/prometherion/Documents/capsule/e2e/missing_tenant_test.go:32  
•
------------------------------ 
creating a Namespace with Tenant selector  
  should be assigned to the selected Tenant
  /home/prometherion/Documents/capsule/e2e/selecting_tenant_test.go:69
STEP: assigning to the Namespace the Capsule Tenant label 
•
------------------------------ 
creating namespaces within a Tenant with resources  
  should contains all replicated resources 
  /home/prometherion/Documents/capsule/e2e/tenant_resources_test.go:181  
STEP: creating the Namespaces  
STEP: checking Limit Range resources 
STEP: checking Network Policy resources 
STEP: checking the Namespace scheduler annotation
STEP: checking the Resource Quota resources
STEP: checking Limit Range resources 
STEP: checking Network Policy resources 
STEP: checking the Namespace scheduler annotation
STEP: checking the Resource Quota resources
STEP: checking Limit Range resources 
STEP: checking Network Policy resources 
STEP: checking the Namespace scheduler annotation
STEP: checking the Resource Quota resources
•
------------------------------ 
when Tenant handles Storage classes  
  should block non allowed Storage Class
  /home/prometherion/Documents/capsule/e2e/storage_class_test.go:61
STEP: non-specifying the class 
STEP: specifying a forbidden class
•
------------------------------ 
when Tenant handles Storage classes  
  should allow enabled Storage Class 
  /home/prometherion/Documents/capsule/e2e/storage_class_test.go:107  
•
------------------------------
changing Tenant managed Kubernetes resources 
  should reapply the original resources upon third party change
  /home/prometherion/Documents/capsule/e2e/tenant_resources_changes_test.go:179
STEP: creating the Namespaces
STEP: changing Limit Range resources
STEP: changing Network Policy resources
STEP: changing Resource Quota resources
STEP: changing Limit Range resources
STEP: changing Network Policy resources
STEP: changing Resource Quota resources
STEP: changing Limit Range resources
STEP: changing Network Policy resources
STEP: changing Resource Quota resources
STEP: changing Limit Range resources
STEP: changing Network Policy resources
STEP: changing Resource Quota resources

• [SLOW TEST:18.366 seconds]
changing Tenant managed Kubernetes resources
/home/prometherion/Documents/capsule/e2e/tenant_resources_changes_test.go:37
  should reapply the original resources upon third party change
  /home/prometherion/Documents/capsule/e2e/tenant_resources_changes_test.go:179
------------------------------
when Tenant handles Ingress classes 
  should block non allowed Ingress class
  /home/prometherion/Documents/capsule/e2e/ingress_class_test.go:64
STEP: non-specifying the class
STEP: using a forbidden class as Annotation
STEP: specifying a forbidden class
•
------------------------------
when Tenant handles Ingress classes 
  should allow enabled Ingress class using the deprecated Annotation
  /home/prometherion/Documents/capsule/e2e/ingress_class_test.go:127
•
------------------------------
when Tenant handles Ingress classes 
  should allow enabled Ingress class using the IngressClassName field
  /home/prometherion/Documents/capsule/e2e/ingress_class_test.go:155

• [SLOW TEST:6.112 seconds]
when Tenant handles Ingress classes
/home/prometherion/Documents/capsule/e2e/ingress_class_test.go:38
  should allow enabled Ingress class using the IngressClassName field
  /home/prometherion/Documents/capsule/e2e/ingress_class_test.go:155
------------------------------
creating a Namespace as Tenant owner 
  should be available in Tenant namespaces list
  /home/prometherion/Documents/capsule/e2e/new_namespace_test.go:53
•STEP: tearing down the test environment


Ran 18 of 18 Specs in 108.157 seconds
SUCCESS! -- 18 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

prometherion:
@MaxFedotov please, can you squash some commits so can we merge this? I'm really thrilled! 🎉

MaxFedotov:
@prometherion done :)

prometherion (approved):
LGTM!

PR #82, opened 2020-09-08
Using HTTP handlers as closures

### LINKED ISSUE #81: Webhook handlers should work as piped middleware
# Describe the feature

It appears we have to replicate or not for some webhook certain business logic.

As an example, we're always [checking if the user impacting the webhook is a Capsule user or not](https://github.com/clastix/capsule/blob/ee0261c069d073fd097dacd7acb59a22396c1b77/pkg/webhook/router.go#L59-L62) but this could be a blocker for future interactions and we should rely on a smart handlers encapsulation, [as HTTP closures](https://www.alexedwards.net/blog/making-and-using-middleware).

1. Separate the Webhook wrapper from the Handlers
2. Define the types to operate with
3. Ensure everything is working fine

# Expected behavior

The test suite must be green upon changes.

prometherion:
Closes #81.

Please,@MaxFedotov review this: I guess this [could be related to your changes](https://github.com/clastix/capsule/pull/71#discussion_r481425844)


MaxFedotov:
@prometherion Thanks a lot for great work!
What I thought is maybe we can use a single hook for both,`networking.k8s.io` and `extensions` groups, as it seems like they had the same logic. We can use a switch on `req.Resource.Group` and find out in which object (either `&networkingv1beta1.Ingress{}` or `&extensionsv1beta1.Ingress{}` to decode it

prometherion:
I'll definitely take a look to this option and prepare an additional commit to achieve this, thanks!

main.go
```diff
@@ -131,10 +132,18 @@ func main() {
...
+		make([]webhook.Webhook, 0),
+		ingress.ExtensionWebhook(utils.InCapsuleGroup(capsuleGroup, ingress.ExtensionHandler(ingress.NewHandlerIngress))),
+		ingress.NetworkingWebhook(utils.InCapsuleGroup(capsuleGroup, ingress.NetworkingHandler(ingress.NewHandlerIngress))),
+		pvc.Webhook(utils.InCapsuleGroup(capsuleGroup, pvc.Handler())),
+		owner_reference.Webhook(utils.InCapsuleGroup(capsuleGroup, owner_reference.Handler())),
+		namespace_quota.Webhook(utils.InCapsuleGroup(capsuleGroup, namespace_quota.Handler())),
+		network_policies.Webhook(utils.InCapsuleGroup(capsuleGroup, network_policies.Handler())),
+		tenant_prefix.Webhook(utils.InCapsuleGroup(capsuleGroup, tenant_prefix.Handler(forceTenantPrefix, protectedNamespaceRegexp))),
+	)
+	err = webhook.Register(mgr, wl...)
```
MaxFedotov, line 146:
```
if err := webhook.Register(mgr, wl...); err != nil {
	setupLog.Error(err, "unable to setup webhooks")
	os.Exit(1)
}
```

pkg/webhook/handler.go
```diff
@@ -23,8 +23,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
+type Func func(ctx context.Context, req admission.Request) admission.Response
+
 type Handler interface {
-	OnCreate(ctx context.Context, req admission.Request, client client.Client, decoder *admission.Decoder) admission.Response
-	OnDelete(ctx context.Context, req admission.Request, client client.Client, decoder *admission.Decoder) admission.Response
-	OnUpdate(ctx context.Context, req admission.Request, client client.Client, decoder *admission.Decoder) admission.Response
+	OnCreate(client client.Client, decoder *admission.Decoder) Func
```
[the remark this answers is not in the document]
prometherion: I was shadowing the type to don't lock in to the underlying implementation since we're not using the Kubebuilder/Operator SDK native implementation, just to avoid breaking changes for the upcoming future in case of signature change.

main.go
prometherion: I guess we have to avoid variable shadowing here, or at least on all the `main.go` file

prometherion:
@MaxFedotov as you suggested, I just implemented with 2f40a35 the single webhook.

If it looks good to you, @bsctl needs to update Capsule Charts too since the Validation Webhook manifest [has changed](https://github.com/clastix/capsule/blob/2f40a3511cdbdc38a479483e02a87c88acd4f920/config/webhook/manifests.yaml#L34-L52)

main.go
prometherion: Resolved with commit 94af699, please review and resolve the conversation if it's ok.

MaxFedotov:
LGTM, thanks a lot 👍

MaxFedotov (approved):

PR #86, opened 2020-09-10
Using pointers and internal handler for Namespace OwnerReferences

### LINKED ISSUE #85: Webhook closures not working
# Bug description

Closures not working in webhooks middleware

# How to reproduce

Steps to reproduce the behaviour:

build capsule from master, load it to kind
try to create namespace as cluster-admin
You will got following error
```
k create ns bla
Error from server (You do not have any Tenant assigned: please, reach out the system administrators): admission webhook "owner.namespace.capsule.clastix.io" denied the request: You do not have any Tenant assigned: please, reach out the system administrators
```

# Expected behavior

Namespace should be created, as kind cluster-admin is not a member of `capsule-user-group`

prometherion:
Closes #85 plus minor `fmt`

### DIFF
--- a/pkg/webhook/owner_reference/patching.go
+++ b/pkg/webhook/owner_reference/patching.go
@@ -43,15 +43,15 @@ func Webhook(handler capsulewebhook.Handler) capsulewebhook.Webhook {
 	return &webhook{handler: handler}
 }
 
-func (w webhook) GetHandler() capsulewebhook.Handler {
-	return &handler{}
+func (w *webhook) GetHandler() capsulewebhook.Handler {
+	return w.handler
 }
 
-func (w webhook) GetName() string {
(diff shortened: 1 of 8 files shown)


pkg/webhook/owner_reference/patching.go
```diff
@@ -43,15 +43,15 @@ func Webhook(handler capsulewebhook.Handler) capsulewebhook.Webhook {
 	return &webhook{handler: handler}
 }
 
-func (w webhook) GetHandler() capsulewebhook.Handler {
-	return &handler{}
+func (w *webhook) GetHandler() capsulewebhook.Handler {
+	return w.handler
```
prometherion, line 47: @MaxFedotov this was the cause of the issue reported on #85: we weren't returning the passed handler but rather the old implementation.

PR #89, opened 2020-09-11
Regexp support for Ingress and Storage classes

### LINKED ISSUE #87: Support for regular expressions in ingressClasses and storageClasses
# Describe the feature

Right now for ingress\storage classes you need to pass an array of distinct values. In our env, when user orders a tenant in k8s cluster, he can specify if he wants dedicated ingress nodes and which ingress he wants (e.g. haproxy\nginx) and then we provision separate nodes with specified ingresses. We want to name these ingress classes using following naming convention: {tenantName}-haproxy, {tenantName-nginx}. So for us, it would be much easier to be able to pass regexp in ingressClasses\storageClasses.

Tenant admin will be able to create tenants with following options (as per discussion with @prometherion):
```
ingressClasses:
  allowed:
  - class-with-no-regex
  - another-class
  allowedRegex: ^oil-.*$
storageClasses:
  allowed:
  - class-with-no-regex
  - another-class
  allowedRegex: ^oil-.*$
```

MaxFedotov:
closes #87


MaxFedotov:
@prometherion please review :)

controllers/tenant_controller.go
```diff
@@ -79,6 +80,22 @@ func (r TenantReconciler) Reconcile(request ctrl.Request) (result ctrl.Result, e
...
+	if len(instance.Spec.IngressClasses.AllowedRegex) > 0 {
+		if _, err := regexp.Compile(instance.Spec.IngressClasses.AllowedRegex); err != nil {
+			r.Log.Info("Unable to compile allowedRegex for ingressClasses", "allowedRegex", instance.Spec.IngressClasses.AllowedRegex)
+			return reconcile.Result{}, nil
+		}
+	}
+
+	// Validate storageClasses regexp
+	if len(instance.Spec.StorageClasses.AllowedRegex) > 0 {
+		if _, err := regexp.Compile(instance.Spec.StorageClasses.AllowedRegex); err != nil {
```
prometherion, line 93: Probably we should move these `regexp.Compile` at a Tenant validation webhook, so the Cluster Admin can easily check if there's a typo in those expressions.

Right now, it's ok here: just let's plan it as Issue/Enhancement.

pkg/webhook/ingress/validating.go
```diff
@@ -120,9 +119,30 @@ func (r *handler) validateIngress(ctx context.Context, c client.Client, object I
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	if !tl.Items[0].Spec.IngressClasses.IsStringInList(*ingressClass) {
+	if len(tl.Items[0].Spec.IngressClasses.AllowedRegex) > 0 || len(tl.Items[0].Spec.IngressClasses.Allowed) > 0 {
+		ingressClassEnforced = true
+	}
+
+	if !ingressClassEnforced {
```
prometherion, line 126: I guess we have to change the behavior, here: we have to cover multiple cases, like denying any kind of Ingress or PVC creation with the following YAML definition:

```
# deny any Ingress
ingressClasses:
  allowed: []
  allowedRegex: ""
```

```
# allow any Ingress
ingressClasses:
  allowed: []
  allowedRegex: .*
```

So all the business logic here should be changed according to this, otherwise would be impossible to achieve also this kind of expected behavior.

pkg/webhook/pvc/errors.go
```diff
@@ -20,24 +20,24 @@ import (
...
 
-func NewValidStorageClassError() error {
-	return &validStorageClassError{}
+func NewStorageClassNotValid() error {
+	return &storageClassNotValid{}
 }
 
-func (validStorageClassError) Error() string {
+func (storageClassNotValid) Error() string {
 	return "A valid Strage Class must be used"
```
prometherion, line 30: noticed a typo here: `Strage`, please can you fix it within your PR?

use_cases.md
```diff
@@ -725,6 +765,8 @@ admission webhook "pvc.capsule.clastix.io" denied the request:
 Storage Class default is forbidden for the current Tenant
 ```
 
+If no `storageClasses` is specified in Tenant spec this policy is not enforced and Alice can use any storage class.
```
prometherion, line 768: According to change requests, that wouldn't be true.

prometherion (changes requested):
Let's make it more agnostic, allowing a _DenyAll_ policy as suggested in the code review.

controllers/tenant_controller.go
MaxFedotov: Let's do it right now and rename `tenant_owner` webhook to `tenant`

pkg/webhook/ingress/validating.go
MaxFedotov: done

pkg/webhook/pvc/errors.go
MaxFedotov: fixed :)

use_cases.md
MaxFedotov: removed

MaxFedotov:
> Let's make it more agnostic, allowing a _DenyAll_ policy as suggested in the code review.

Done, please review. If everything is OK - I will squash commits and force push

pkg/webhook/ingress/validating.go
```diff
@@ -108,9 +109,11 @@ func (r *handler) ingressFromRequest(req admission.Request, decoder *admission.D
 }
 
 func (r *handler) validateIngress(ctx context.Context, c client.Client, object Ingress) admission.Response {
+	var valid, matched bool
 	ingressClass := object.IngressClass()
+
 	if ingressClass == nil {
-		return admission.Errored(http.StatusBadRequest, fmt.Errorf("A valid Ingress Class must be used"))
+		return admission.Errored(http.StatusBadRequest, NewIngressClassForbidden(*ingressClass))
```
prometherion, line 116: Oh, thanks: didn't notice a custom error was missing! <3

pkg/webhook/tenant/validating.go
```diff
@@ -68,6 +68,21 @@ func (r *handler) OnCreate(client client.Client, decoder *admission.Decoder) cap
 		if !matched {
 			return admission.Denied("Tenant name has forbidden characters")
 		}
+
+		// Validate ingressClasses regexp
```
prometherion, line 72: You so kind @MaxFedotov , thanks! <3

prometherion (approved):
What an impressive job @MaxFedotov, LGTM!

Please, proceed with squash so we can get it merged.

PR #104, opened 2020-10-04
Adding support for RegistryClasses

### LINKED ISSUE #1: Ensure all images come from a trusted registry declared in the Tenant
# Describe the feature

A Tenant could just use a specific set of images from a trusted registry.

Should we use a list of allowed registries?

1. setup a Validating Admission webhook for every Pod
2. retrieve the Tenant from the namespace indexer
3. check if the image starts with the desired registry name

# Expected behavior

- if the Pod is running on a non-Capsule namespace, return `Allowed` without further actions
- if the Pod is running on a Capsule namespace with a non allowed registry, return `Forbidden`

AdheipSingh:
Fixes #1. 

Will allow running declared container images from the specified registries defined in the _Tenant_ spec.

### DIFF
--- a/api/v1alpha1/registry_class_list.go
+++ b/api/v1alpha1/registry_class_list.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2020 Clastix Labs.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"sort"
+	"strings"
+)
+
+type RegistryList []string
+
+func (n RegistryList) Len() int {
+	return len(n)
+}
+
+func (n RegistryList) Swap(i, j int) {
+	n[i], n[j] = n[j], n[i]
+}
+
+func (n RegistryList) Less(i, j int) bool {
+	return strings.ToLower(n[i]) < strings.ToLower(n[j])
+}
+
+func (n RegistryList) IsStringInList(value string) (ok bool) {
+	sort.Sort(n)
+	i := sort.SearchStrings(n, value)
+	ok = i < n.Len() && n[i] == value
+	return
+}
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -130,6 +130,23 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
+      path: /validating-v1-registry
+  failurePolicy: Fail
+  name: pod.capsule.clastix.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    resources:
+    - pods
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
       path: /validating-v1-tenant
   failurePolicy: Fail
   name: tenant.capsule.clastix.io
--- a/main.go
+++ b/main.go
@@ -43,6 +43,7 @@ import (
 	"github.com/clastix/capsule/pkg/webhook/network_policies"
 	"github.com/clastix/capsule/pkg/webhook/owner_reference"
 	"github.com/clastix/capsule/pkg/webhook/pvc"
+	"github.com/clastix/capsule/pkg/webhook/registry"
 	"github.com/clastix/capsule/pkg/webhook/service_labels"
 	"github.com/clastix/capsule/pkg/webhook/tenant"
 	"github.com/clastix/capsule/pkg/webhook/tenant_prefix"
@@ -146,6 +147,7 @@ func main() {
 		make([]webhook.Webhook, 0),
 		ingress.Webhook(utils.InCapsuleGroup(capsuleGroup, ingress.Handler())),
 		pvc.Webhook(utils.InCapsuleGroup(capsuleGroup, pvc.Handler())),
+		registry.Webhook(utils.InCapsuleGroup(capsuleGroup, registry.Handler())),
 		owner_reference.Webhook(utils.InCapsuleGroup(capsuleGroup, owner_reference.Handler(forceTenantPrefix))),
 		namespace_quota.Webhook(utils.InCapsuleGroup(capsuleGroup, namespace_quota.Handler())),
 		network_policies.Webhook(utils.InCapsuleGroup(capsuleGroup, network_policies.Handler())),
--- a/pkg/webhook/registry/validating.go
+++ b/pkg/webhook/registry/validating.go
@@ -1,0 +1,118 @@
+/*
+Copyright 2020 Clastix Labs.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registry
+
+import (
+	"context"
+	"net/http"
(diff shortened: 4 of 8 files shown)


config/webhook/manifests.yaml
```diff
@@ -125,6 +125,23 @@ webhooks:
     - CREATE
     resources:
     - persistentvolumeclaims
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validating-v1-registry
+  failurePolicy: Fail
```
prometherion, line 134: We cannot set the failurePolicy to Fail otherwise we could end up in a recursion loop.

Let's imagine the installation of Capsule along with this Validating Webhook: the Capsule Pod would need to pass through this but that wouldn't be yet ready since hasn't started yet.

It's something we already had with the Service resource annotations/labels webhook made by @maxfedotov

main.go
```diff
@@ -146,6 +147,7 @@ func main() {
 		make([]webhook.Webhook, 0),
 		ingress.Webhook(utils.InCapsuleGroup(capsuleGroup, ingress.Handler())),
 		pvc.Webhook(utils.InCapsuleGroup(capsuleGroup, pvc.Handler())),
```
prometherion, line 149: Do we need the `InCapsuleGroup` Middleware?

I guess a Pod would be started by a ReplicaSet or a DaemonSet and this would be ignored.

pkg/webhook/registry/validating.go
```diff
@@ -0,0 +1,118 @@
...
+		var valid, matched bool
+		pod := &v1.Pod{}
+
+		if err := decoder.Decode(req, pod); err != nil {
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+
+		containers := pod.Spec.Containers
+
+		for _, container := range containers {
```
prometherion, line 73: Why iterating over Containers checking if image is not empty?

By default it's a mandatory field, so I don't get what we have to perform this kind of check.

pkg/webhook/registry/validating.go
```diff
@@ -0,0 +1,118 @@
...
+
+		if len(tl.Items[0].Spec.RegistryClasses.Allowed) > 0 {
+			for _, container := range containers {
+				valid = tl.Items[0].Spec.RegistryClasses.Allowed.IsStringInList(container.Image)
+			}
+		}
+
+		if len(tl.Items[0].Spec.RegistryClasses.Allowed) > 0 {
+			for _, container := range containers {
+				matched, _ = regexp.MatchString(tl.Items[0].Spec.RegistryClasses.AllowedRegex, container.Image)
```
prometherion, line 94: Why not putting this inside the for cycle defined at line 87?

pkg/webhook/registry/validating.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+		}
+
+		if !valid && !matched {
+			for _, container := range containers {
```
prometherion, line 99: I guess we can easily avoid to iterate over all the containers, here.

If not valid and not matched, let's disallow it.

prometherion (commented):
@AdheipSingh please, can you provide further details upon my comments?

config/webhook/manifests.yaml
AdheipSingh: makes sense, will change to ignore

pkg/webhook/registry/validating.go
AdheipSingh: A pod can have multiple containers each having multiple images, so the webhook must check each image in case they are from a trusted registery. Thats why we need to range over all containers

pkg/webhook/registry/validating.go
AdheipSingh: sure can be done :)

pkg/webhook/registry/validating.go
AdheipSingh: same reason 
> A pod can have multiple containers each having multiple images, so the webhook must check each image in case they are from a trusted registery. Thats why we need to range over all containers

prometherion: There's no need to iterate here: we can safely return the admission error without iteration, here.

pkg/webhook/registry/validating.go
prometherion: What I'm trying to say is that the Pod image is a mandatory field and this check is already performed at API Level (or at client one, depending on the option).

We can skip this kind of check since we're getting already passed data by the API server.

AdheipSingh: ah i got it now..... agreed

pkg/webhook/registry/validating.go
AdheipSingh: In case i remove range, is this the error msg you want to go with ? ```NewregistryClassForbidden(pod.Spec.Containers[0].Image)```

AdheipSingh: BTW can we use admissioned.denied here ?

prometherion: Would be nice adding the list of container images that are not allowed to run using the `admission.Denied`, something like:

```
The Pod cannot be created due to following image(s) not allowed: ${images...}
```

If it's too difficult to collect a slice of unmatched and non valid images, let's ignore this: let's focus on addressing the feature.

pkg/webhook/registry/validating.go
```diff
@@ -0,0 +1,118 @@
...
+		tl := &capsulev1alpha1.TenantList{}
+		if err := c.List(ctx, tl, client.MatchingFieldsSelector{
+			Selector: fields.OneTermEqualSelector(".status.namespaces", pod.Namespace),
+		}); err != nil {
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+
+		if len(tl.Items[0].Spec.RegistryClasses.Allowed) > 0 {
+			for _, container := range containers {
+				valid = tl.Items[0].Spec.RegistryClasses.Allowed.IsStringInList(container.Image)
```
prometherion, line 88: I guess this kind of check is broken: follow the said example.

I got a Tenant with the this `registryPolicy`

```yaml
kind: Tenant
apiGroup: capsule.clastix.io/vealpha1
spec:
  registryClasses:
    allowed:
      - "hub.my-org.tld"
    allowedRegex: ""
```

Let's imagine I got a PodTemplate made of two containers:

```yaml
containers:
- name: denied
  image: hub.docker.io/alpine:latest
- name: allowed
  image: hub.my-org.tld/repository/app:latest
```

According to the said code, iterating over this array, with the `denied` container image (`hub.docker.io/alpine:latest`) the variable `valid` will be set to `false`; with the next value, provided by container `allowed` running the image `hub.my-org.tld/repository/app:latest`, the `valid` variable will be set to `true` and this is not the expected behavior, since the registry policy should deny any non allowed or non-valid container registry.

The same applies also for the `matched` code at line 94.

prometherion (changes requested):
There's a typo in the business logic that should be addressed.

api/v1alpha1/registry_class_list.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (n RegistryList) Swap(i, j int) {
+	n[i], n[j] = n[j], n[i]
+}
+
+func (n RegistryList) Less(i, j int) bool {
+	return strings.ToLower(n[i]) < strings.ToLower(n[j])
+}
+
+func (n RegistryList) IsStringInList(value string) (ok bool) {
```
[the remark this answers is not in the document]
AdheipSingh: ok with the example its more clear. 
We will still need to get to each of the container to get the image string and then do parsing. 
```registry_name/repo_name/image:tag```

pkg/webhook/registry/validating.go
AdheipSingh: In case one image is valid and other is not, we still go with Valid set to False ?

prometherion: Exactly, all the containers images must adhere to the registry policy.

api/v1alpha1/registry_class_list.go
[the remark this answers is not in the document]
bsctl: @prometherion @AdheipSingh If you don't specify a registry hostname, Kubernetes assumes that you mean the Docker public registry `hub.docker.io`. If you want to use a different registry for a given pod, you have to specify it in pod manifest. If you want to change the general behavior, e.g. use a private registry by default, you should to change it at nodes level, pls see [here](https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry). Hope this helps.

prometherion:
Hi @AdheipSingh, any news about this PR?

May I help you with something?

AdheipSingh:
@prometherion
 i haven't got time to get this finished, not sure when. In case this priority, please feel to take this.

prometherion:
Closing for #142.

PR #116, opened 2020-10-30
fix endpoint creation for services with selectors when servicesMetadata is specified for tenant

### LINKED ISSUE #115: Endpoints are not created for services with selectors if tenants servicesMetadata is specified
# Bug description

Enpoints are not created if `servicesMetadata.additionalLabels` or `servicesMetadata.additionalLabels` are used for tenant and service is created with selector (for headless services everything is OK)

# How to reproduce

Create new tenant with:
```
  servicesMetadata:
    additionalAnnotations: null
    additionalLabels:
      capsule.clastix.io/tenant: system
```
Create service with selector:
```
apiVersion: v1
kind: Service
metadata:
  name: kube-node-provisioner-autoscaler
spec:
  type: ClusterIP
  ports:
    - port: 8086
      targetPort: 8086
      protocol: TCP
      name: autoscaler-http
  selector:
    app.kubernetes.io/name: kube-node-provisioner
```

service will be created, but endpoints not

# Expected behavior

endpoints should be created

# Logs

In controller-manager logs you can find following:
```
I1030 14:22:52.501562       1 event.go:291] "Event occurred" object="kube-node-provisioner-autoscaler" kind="Endpoints" apiVersion="v1" type="Warning" reason="FailedToCreateEndpoint" message="Failed to create endpoint for service kube-node-provisioner-system/kube-node-provisioner-autoscaler: admission webhook \"service.labels.capsule.clastix.io\" denied the request: Namespace \"\" not found"
```

That is because when endpoint is created automatically for service with selector, it doesn't have Namespace in it's spec.

Same error, but added some debug info
```
I1030 17:41:04.345685       1 event.go:291] "Event occurred" object="kube-node-provisioner-autoscaler" kind="Endpoints" apiVersion="v1" type="Warning" reason="FailedToCreateEndpoint" message="Failed to create endpoint for service kube-node-provisioner-system/kube-node-provisioner-autoscaler: admission webhook \"service.labels.capsule.clastix.io\" denied the request: unable to process obj: &Endpoints{ObjectMeta:{kube-node-provisioner-autoscaler      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[app.kubernetes.io/instance:kube-node-provisioner app.kubernetes.io/managed-by:Helm app.kubernetes.io/name:kube-node-provisioner app.kubernetes.io/version:0.0.4 capsule.clastix.io/tenant:system helm.sh/chart:kube-node-provisioner-0.0.4 provisioner.k8s.io/app:kube-node-provisioner] map[] [] []  [{kube-controller-manager Update v1 2020-10-30 17:41:14 +0000 UTC FieldsV1 {\"f:metadata\":{\"f:labels\":{\".\":{},\"f:app.kubernetes.io/instance\":{},\"f:app.kubernetes.io/managed-by\":{},\"f:app.kubernetes.io/name\":{},\"f:app.kubernetes.io/version\":{},\"f:capsule.clastix.io/tenant\":{},\"f:helm.sh/chart\":{},\"f:provisioner.k8s.io/app\":{}}},\"f:subsets\":{}}}]},Subsets:[]EndpointSubset{EndpointSubset{Addresses:[]EndpointAddress{},NotReadyAddresses:[]EndpointAddress{EndpointAddress{IP:10.244.0.6,TargetRef:&ObjectReference{Kind:Pod,Namespace:kube-node-provisioner-system,Name:kube-node-provisioner-7f5c8b677-lczhz,UID:4779faed-6ad1-4963-ad91-4f50def9be15,APIVersion:,ResourceVersion:1586,FieldPath:,},Hostname:,NodeName:*capsule-control-plane,},},Ports:[]EndpointPort{EndpointPort{Name:autoscaler-http,Port:8086,Protocol:TCP,AppProtocol:nil,},},},},}"
```

As you can see, there is no `Namespace` field in `ObjectMeta`

But there is `Namespace` field in `TargetRef`, so in order to fix this bug, i will add following to 
https://github.com/clastix/capsule/blob/ee6e3aa0df79ada668d68f3a174eb09ae2617bc5/pkg/webhook/service_labels/types.go#L50-L52

```
func (ep Endpoints) Namespace() string {
	namespace := ep.GetNamespace()
	// For ep, which are created automatically using service selector namespace will always be empty, so we had to take it from TargetRef
	if namespace == "" {
		namespace = ep.Subsets[0].Addresses[0].TargetRef.Namespace
	}
	return namespace
}
```

MaxFedotov:
fixed #115

### DIFF
--- a/pkg/webhook/service_labels/types.go
+++ b/pkg/webhook/service_labels/types.go
@@ -47,8 +47,13 @@ type Endpoints struct {
 	*corev1.Endpoints
 }
 
-func (ep Endpoints) Namespace() string {
-	return ep.GetNamespace()
+func (ep Endpoints) Namespace() (namespace string) {
+	namespace = ep.GetNamespace()
+	// For ep, which are created automatically using service selector namespace will always be empty, so we had to take it from TargetRef
+	if len(namespace) == 0 {
+		namespace = ep.Subsets[0].Addresses[0].TargetRef.Namespace
+	}
+	return
 }
 
 func (ep Endpoints) Labels() map[string]string {


prometherion:
Failed test sounds flaky, going to push it back.

pkg/webhook/service_labels/types.go
```diff
@@ -48,7 +48,12 @@ type Endpoints struct {
 }
 
 func (ep Endpoints) Namespace() string {
-	return ep.GetNamespace()
+	namespace := ep.GetNamespace()
```
prometherion, line 51: maybe let's use named returns to avoid allocation of a new variable, here?

```go
func (ep Endpoints) Namespace() (namespace string) {
  namespace = ep.GetNamespace()
  if len(namespace) == 0 {
    namespace = ep.Subsets[0].Addresses[0].TargetRef.Namespace
  }
  return
}
```

prometherion (approved):
There's just a concern regarding a matter of style: BTW, if tests are green, LGTM.

pkg/webhook/service_labels/types.go
MaxFedotov: Ok, applied and rebased :)

prometherion:
@MaxFedotov I just merged #121 and we should address the flakiness of the Kubernetes matrix: could you rebase so we can test it and get it green?

Issue #113, opened 2020-10-29
Tenant names with dashes are denied by admission webhook


d-m:
# Bug description

Tenant names with dashes fail admission we hook with "Tenant name has forbidden characters." Do dashes cause an issue or is this solely due to the regex used?

# How to reproduce

1. Attempt to creat a Tenant resource with name "test-tenant"

# Expected behavior

A tenant with the name "test-tenant" is created. 

# Logs

```
Error: admission webhook "tenant.capsule.clastix.io" denied the request: Tenant name has forbidden characters
```

prometherion:
It seems we have a regex, there:

https://github.com/clastix/capsule/blob/ee6e3aa0df79ada668d68f3a174eb09ae2617bc5/pkg/webhook/tenant/validating.go#L67-L70

I recall a discussion with @MaxFedotov that implemented that, also according to `git blame`, but honestly I don't recall what's the downside of having dashes on Tenant name. Pretty sure, the Tenant name is used across referenced resources to allow selection, but we just need to adhere to DNS RFC-1123.

Please, Max, would you mind providing a reminder behind this designing decision?

MaxFedotov:
Hi, 
this was related to https://github.com/clastix/capsule/issues/76#issuecomment-688826105
Without this restrictions, we are unable to distinguish namespace name from tenant name

d-m:
Ah I see. This may cause some issues for my use case. I'm trying to map Tenant resources to some exists groups I have, but some of these groups have dashes. I'd like to keep them if possible so that the namespace prefixes make sense to the users, since they're familiar with the group names they're in. 

Unfortunately, I can't drop the dashes for the Tenant name since I sometimes have other groups that would then conflict. For example, unique groups "abc" and "ab-c".

prometherion:
I think we can try to make an effort and solve this problem from the upside-down, rather.

I just opened a PR and would be nice having feedback from @MaxFedotov since I guess we can easily implement it when test suite is green! :)

prometherion:
Good news @d-m: with the latest contributions Capsule now accepts also Tenant names with dashes as DNS RFC-1123 🎉 

Please, give it a try building Capsule from `master` and feel free to open back this issue in the case is not working as expected.

Issue #106, opened 2020-10-14
Provide a more meaningful error message when `--force-tenant-prefix` is used and the validation webhook rejects the request


bsctl:
# Bug description
Provide a more meaningful error message when `--force-tenant-prefix` is used and the validation webhook rejects the request.

# How to reproduce
1. Start capsule with `--force-tenant-prefix` option
2. Create a tenant
3. Login as tenant owner
4. Create a namespace without the tenant prefix: the validation webook denies and returns a generic message "`admission webhook "owner.namespace.capsule.clastix.io" denied the request: Tenant.capsule.clastix.io "tontolo" not found`"

# Expected behavior
From a user experience perspective, we should provide a more meaningful message informing the user why it failed and how to fix it. For example: "`admission webhook "owner.namespace.capsule.clastix.io" denied the request: the namespace "tontolo" is not valid. It should be <tenant_name>-tontolo`" or similar.

# Logs
na

# Additional context

- Capsule version: v0.0.1
- Kubernetes version: v1.19.2

bsctl:
@prometherion trying to understand the code base, we see two different mutating web hooks trying to enforce the `--force-tenant-prefix` policy:

- https://github.com/clastix/capsule/blob/0ce901f8dd5b6559c9dcca7036a5efa6db10f50f/pkg/webhook/tenant_prefix/patching.go#L70

- https://github.com/clastix/capsule/blob/0ce901f8dd5b6559c9dcca7036a5efa6db10f50f/pkg/webhook/owner_reference/patching.go#L71

Not sure to understand the rationale behind it. Could you please clarify?

bsctl:
We found two different behaviours trying to create a namespace (as tenant owner) that is not compliant the `--force-tenant-prefix` policy

First case:
```
$ kubectl create ns foo
Error from server: admission webhook "owner.namespace.capsule.clastix.io" denied the request: Tenant.capsule.clastix.io "foo" not found
```
in the case above, the namespace `foo` is evaluated as tenant by admission webhook `owner.namespace.capsule.clastix.io`

Second case:

```
$ cat <<EOF >./foo-ns.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: foo
  labels:
    capsule.clastix.io/tenant: gas
EOF

$ kubectl create -f foo-ns.yaml
[...]

PR #118, opened 2020-10-31
Allowing dash on Tenant namespace

prometherion:
This is an improvement started from the discussion on #113 and addressing also #106.

We would like to be less strict regarding Tenant naming, allowing end-users to use dashes according to [DNS RFC-1123](https://tools.ietf.org/html/rfc1123), the same applied by Kubernetes for some resources.

I'd like to get a review from @MaxFedotov since he was the author of the force tenant prefix.

### DIFF
--- a/e2e/custom_capsule_group_test.go
+++ b/e2e/custom_capsule_group_test.go
@@ -32,7 +32,7 @@ import (
 var _ = Describe("creating a Namespace as Tenant owner with custom --capsule-group", func() {
 	tnt := &v1alpha1.Tenant{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "tenantassignedcustomgroup",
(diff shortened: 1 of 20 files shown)


prometherion:
Tests are flaky and must be fixed, tested changes on `1.18.0`:

```
Ran 29 of 29 Specs in 243.480 seconds
SUCCESS! -- 29 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

prometherion:
Same on `1.16.15`

```
Ran 27 of 29 Specs in 224.450 seconds
SUCCESS! -- 27 Passed | 0 Failed | 0 Pending | 2 Skipped
PASS
```

MaxFedotov:
lgtm!

PR #121, opened 2020-10-31
Making e2e tests less flaky

### LINKED ISSUE #120: Refactoring e2e test suite
It appears, from time to time, we got flaky tests as happened [here](https://github.com/clastix/capsule/pull/118/checks?check_run_id=1335492495): in this case, is not definitely an issue of an underlying Capsule error, but rather the respawn of the Capsule pod not so quick to deal with the args changes.

My suggestion is to use the `Eventually` helper also for the resource creations inside the `JustBeforeEach` blocks.

https://github.com/clastix/capsule/blob/ee6e3aa0df79ada668d68f3a174eb09ae2617bc5/e2e/custom_capsule_group_test.go#L52-L55

Last but not least, we could also refactor the test utilities, returning an `AsyncAssertion` rather than a `Should` assertion: this would help also the error reporting since the assertion would be made at the test feature implementation.

prometherion:
Closes #120.


MaxFedotov:
@prometherion looks superb!

PR #129, opened 2020-11-04
Enforcing back tenant selection using closest match for namespace prefix

### LINKED ISSUE #123: Assign a namespace to the tenant from the tenant prefix used in the namespace
# Describe the feature
When `--force-tenant-prefix` is set, multiple tenants are assigned to the same owner, and the user creates a new namespace, capsule should infer the tenant to assign from the tenant prefix used in the namespace. Before the recent rework introduced by #118, this was the behaviour. 

Currently, when multiple tenants are assigned to the same owner, the user must specify the tenant with the label `capsule.clastix.io/tenant=<desired_tenant>` otherwise capsule denies with the following message `Unable to assign namespace to tenant. Please use capsule.clastix.io/tenant label when creating a namespace`.

From the UX point of view, the new behaviour looks worse. A user having more than one tenant is expecting the system to assign the namespace to the right tenant because the user is following the rule that the system is requiring. He should not be forced to add a label in the namespace to indicate the tenant he wants to be assigned. The tenant is already indicated into the name of the namespace. Forcing the user to specify it again, it looks a poor UX.

It would be nice to revert this useful feature for a better user experience.

1. the capsule controller is started with `--force-tenant-prefix` by the cluster admin
2. the tenant owner gets assigned more than one tenant, eg. `oil` and `gas`
3. the tenant owner is aware of the tenant prefix rule, so she just creates a new namespace `oil-production`
4. capsule infers the desired tenant from the namespace prefix and assign `oil-production` to the `oil` tenant

# Expected behavior
When `--force-tenant-prefix` is set, Capsule infers the desired tenant from the namespace prefix and assign a new namespace to the right tenant

prometherion:
Should close #123.

### DIFF
--- a/e2e/force_tenant_prefix_test.go
+++ b/e2e/force_tenant_prefix_test.go
@@ -32,7 +32,26 @@ import (
 var _ = Describe("creating a Namespace with --force-tenant-name flag", func() {
 	t1 := &v1alpha1.Tenant{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "first",
+			Name: "awesome",
+		},
+		Spec: v1alpha1.TenantSpec{
+			Owner: v1alpha1.OwnerSpec{
+				Name: "john",
+				Kind: "User",
+			},
+			NamespacesMetadata: v1alpha1.AdditionalMetadata{},
+			ServicesMetadata:   v1alpha1.AdditionalMetadata{},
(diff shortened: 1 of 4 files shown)


prometherion:
@MaxFedotov please, take a look at the updated test case: I guess it's safe enough since we're implementing a sort of closest match.

MaxFedotov:
@prometherion LGTM! Thank!

Issue #112, opened 2020-10-24
Leave the tenant owners to handle namespaced CRDs in their own namespaces


bsctl:
# Describe the feature
Leave the tenant owner appropriate RBAC privileges to handle namespaced CRDs in the owned namespaces. Currently, we assign the [pre-defined](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) `admin` ClusterRole to the tenant owner, limited to its namespaces only.

However, this role prevents the tenant owner to handle, e.g. create/get/list/patch/delete namespaced custom resources. For example, the owner cannot create `CI/CD` applications and projects, typically defined as custom resources:

```
$ kubectl auth --as=alice --as-group=capsule.clastix.io can-i get applications.argoproj.io -n oil-utils
no
$ kubectl auth --as=alice --as-group=capsule.clastix.io can-i get appprojects.argoproj.io -n oil-utils
no
```

# What would the new user story look like?
Two options come in my mind:

1. We can define a new ClusterRole for all the Tenant owners, eg. `capsule:tenant-admin`, to use instead of the pre-defined `admin`. This ClusteRole will have permissions `*/*` so any custom resource API is permitted for any Tenant. This may be not the best option.

2. We can add another field in the Tenant manifest where to specify, with more safe granularity, the additional permissions per Tenant. For example, in the tenant `oil` manifest:

```
apiVersion: capsule.clastix.io/v1alpha1
kind: Tenant
metadata:
  name: oil
spec:
  additionalAuthzRules:
  - apiGroups:
     - argoproj.io
    resources:
    - applications
    - appprojects
    verbs:
     - create
     - get
     - list
     - watch
     - update
     - patch
     - delete
```

When the tenant above is created, the Capsule controller generates "aggregate" ClusterRoles from the specs above, eg.:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: aggregate-oil
aggregationRule:
  clusterRoleSelectors:
  - matchLabels:
      capsule.calstix.io/aggregate-oil: "true"
rules: [] 
```

and

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: capsule-tenant:argoproj.io
  labels:
    capsule.calstix.io/aggregate-oil: "true"
rules:
- apiGroups:
  - argoproj.io
  resources:
  - applications
  - appprojects
  verbs:
  - create
  - get
  - list
  - watch
  - update
[...]

PR #133, opened 2020-11-13
Supporting additional Role Bindings per Tenant

### LINKED ISSUE #52: Dedicated PodSecurityPolicy per Tenant
# Describe the feature
It would be nice to let the cluster admin to set a dedicated Pod Security Policy (PSP) per tenant. This is likely to be a requirement in a multi-tenancy environment.

A PSP is defined at cluster scope. To assign the PSP to users/serviceaccounts in a tenant we can use RoleBindings for each namespace created in the tenant. The Capsule should enforce the creation of RoleBinding assigning the PSP.

1. The cluster admin creates a PSP for a given tenant:

```yaml
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: psp:restricted
spec:
  privileged: false
  # Required to prevent escalations to root.
  ...
```

2. then creates a ClusterRole to assign the PSP

```yaml
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: psp:restricted
rules:
- apiGroups:
  - extensions
  resources:
  - podsecuritypolicies
  resourceNames:
  - restricted # the psp we are giving access to
  verbs:
  - use
```

3. the cluster admin assign such PSP to the tenant:

```yaml
  kind: Tenant
  metadata:
      name: oil
  spec:
    podSecurityPolicies:
    - psp:restricted
 ...
```

# Expected behavior
For each namespace created into the tenant, the Capsule controller creates a RoleBinding assigning the PSP to all the services accounts in such namespace:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: psp:restricted
  namespace:
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: psp:restricted
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:serviceaccounts
```

prometherion:
Closes #52, although it aims to solve a bigger problem: the additional field `additionalRoleBindings` allows to bind RBAC for PSP and CRDs at Tenant Namespaces level (#112), too.

Two birds, one stone.

This change is useful also for PSP integration: documenting how to define basic PSP to run Capsule with this admission controller.


prometherion:
Failed stages are flaky, can be safely ignored.

bsctl (commented):
Tested on both the scenarios:

- PSPs
- CRDs

and it works very good. LGTM

One more question: the tenant owner (alice) can delete RoleBindings from her namespaces. Then Capsule recreates them after a while and this is as expected. However, I noted that for other Capsule-enforced resources, i.e. NetworkPolicies, LimitRanges, there is a validating webhook that prevents the user to delete the resource. Is there any reason of this different behaviour?

prometherion:
> One more question: the tenant owner (alice) can delete RoleBindings from her namespaces. Then Capsule recreates them immediately and this is as expected. However, I noted that for other Capsule-enforced resources, i.e. NetworkPolicies, LimitRanges, there is a validating webhook that prevents the user to delete the resource. Is there any reason for this different behaviour?

If a Tenant Owner is going to delete additional Role Bindings is self-injuring itself, just for a tiny time windows: that means no PSP applied or capabilities for CRDs.

With Limit Range and Network Policy resources it's totally different: deleting them and using the tiny interval to deploy a container with huge amount of CPU, Memory or storage request could lead to a DoS, and Network Policy can provide another small time frame to establish a persistent connection and hijacking the system.

So I guess there's no need of webhook here.

PR #135, opened 2020-11-16
Bumping Makefile to v0.0.2 and minor hotfix on Kustomize

### LINKED ISSUE #134: The docker image is built with wrong tag version
# Bug description
When building from master branch with `make docker-build` the produced artefact is labeled as:

```
Successfully built 4482ce7dd7aa
Successfully tagged quay.io/clastix/capsule:0.0.1 
```

Currently the image should be tagged as `latest`

# How to reproduce

```
$ git pull https://github.com/clastix/capsule
$ cd capsule
$ make docker-build
$ docker images |grep capsule
```

# Expected behavior
Tag image tagged with latest tag.

# Logs

If applicable, please provide logs of `capsule`.

In a standard stand-alone installation of Capsule,
you'd get this by running `kubectl -n capsule-system logs deploy/capsule-controller-manager`.

# Additional context

- Capsule version: (`capsule --version`)
- Kubernetes version: (`kubectl version`)

prometherion:
Closes #134

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,9 @@ jobs:
         k8s-version: ['v1.16.15', 'v1.17.11', 'v1.18.8', 'v1.19.1']
     runs-on: ubuntu-18.04
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Cache Go modules and Docker images
       uses: actions/cache@v1
       env:
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,15 +11,25 @@ RUN go mod download
 
 # Copy the go source
 COPY main.go main.go
+COPY version.go version.go
 COPY api/ api/
 COPY controllers/ controllers/
 COPY pkg/ pkg/
-COPY version/ version/
-
-ARG VERSION
+COPY .git .git
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -ldflags "-X github.com/clastix/capsule/version.Version=${VERSION}" -o manager main.go
+RUN git config --get remote.origin.url > /tmp/GIT_REPO && \
+    git rev-parse --short HEAD > /tmp/GIT_HEAD_COMMIT && \
+    git describe --abbrev=0 --tags > /tmp/GIT_LAST_TAG && \
+    git rev-parse --short $(cat /tmp/GIT_LAST_TAG) > /tmp/GIT_TAG_COMMIT && \
+    git diff $(cat /tmp/GIT_HEAD_COMMIT) $(cat /tmp/GIT_TAG_COMMIT) --quiet > /tmp/GIT_MODIFIED1 || echo '.dev' > /tmp/GIT_MODIFIED1 && \
+    git diff --quiet > /tmp/GIT_MODIFIED2 || echo '.dirty' > /tmp/GIT_MODIFIED2 && \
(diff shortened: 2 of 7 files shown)


bsctl (commented):
> `latest` is an anti-pattern and should be avoided at all since it's mutable

This is not the case, imho. The image built from master branch is **actually** the `latest` artefact available for the users. It is **actually different** from the official release stored on the public repo. Having both the images (one from repo and another one from make build) labeled with the same tag will confuse the users.

> Building from master should produce a build with the upcoming patch release with a suffix of the commit SHA

Do you mean something like `v0.0.2-xyz`?

prometherion:
I think we should adhere to the pattern that Kubernetes is sharing, and, e.g., you can notice there's no `latest` on the [API Server](https://console.cloud.google.com/gcr/images/google-containers/GLOBAL/kube-apiserver?gcrImageListsize=30&gcrImageListquery=%255B%257B_22k_22_3A_22_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22latest_5C_22_22%257D%255D) image hosted on gcr.io.

> Do you mean something like v0.0.2-xyz?

We could do something as `v0.0.3-qwerty` where `0.0.3` is the patch release increased by one and `qwerty` the git commit SHA, but as I said, that's out of our control here: we could think something for it when releasing release candidate.

bsctl:
> I think we should adhere to the pattern that Kubernetes is sharing, and, e.g., you can notice there's no latest on the API Server image hosted on gcr.io.

We do not have to push the tag `latest` on the official public repo and, yes, this is the anti-pattern you're referring to.

My point is that when you build from master branch, you're locally building on your workstation something that is the "state of the art" of the project in a given moment of the day and it can be different from the image released on the public repo. The image you build from master can be a different thing many times a day (depending on our commit rate on master branch) and I think it is not correct to label it as the same as the frozen release image on the public repo.

It's user's responsibility to avoid to deploy the tag `latest` in production avoiding the anti-pattern above.

prometherion:
I get your point, but we're hitting a real edge-case and I'd say it's safe enough, right now.

bsctl:
@prometherion ok let's to take the PR approach for now but I expect troubles in the future because of that.

prometherion:
Please @bsctl, take a review: we're getting some improvements here and there.

First, we're enhancing the version retrieval: according to the latest changes, we're going to have the following lines.

```json
{"level":"info","ts":1605604746.9184825,"logger":"setup","msg":"Capsule Version v0.0.2 d533f66.dev.dirty"}
{"level":"info","ts":1605604746.9185152,"logger":"setup","msg":"Build from: git@github.com:clastix/capsule.git"}
{"level":"info","ts":1605604746.9185236,"logger":"setup","msg":"Build date: 2020-11-17T09:18:13"}
{"level":"info","ts":1605604746.9185975,"logger":"setup","msg":"Go Version: go1.13.14"}
{"level":"info","ts":1605604746.9186049,"logger":"setup","msg":"Go OS/Arch: linux/amd64"}
```

As you can see, we're getting the latest versioned tag (`v0.0.2`) along with the current commit sha (`d533f66`).

The `dev` part informs you're running outside the specified version with `dirty` changes, so uncommitted ones.

Regarding the tag name, I'd like to keep the latest tag: I understand your point of view but Capsule, from my perspective, should not be installed from source code.

We're providing the way to perform this also without the need of Helm or bare YAML files but it's more for development stuff, and developers are intended to know what they're doing.

MaxFedotov:
@prometherion 
btw, I had one thought about improving logging. This thing with `ts` is very hard to troubleshoot in logs

prometherion:
@MaxFedotov maybe let's raise this discussion in another issue so we can evaluate it?

bsctl:
@prometherion lgtm

PR #138, opened 2020-11-17
Providing additional print column for the nodeSelector Tenant spec

### LINKED ISSUE #137: Add the node pool label as additionalPrinterColumns
# Describe the feature
It would be nice to have the node pool label of a Tenant as Additional Printer Column in the output of `kubectl get tenants`

Having a tenant oil with a given node selector

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  nodeSelector:
    pool: caas
...
```

you get

```
$ kubectl get tenants

NAME      NAMESPACE QUOTA   NAMESPACE COUNT   OWNER NAME   OWNER KIND   SELECTOR   AGE
oil      10               3                 alice          User         pool=caas         39d
```

# Expected behavior
see above

prometherion:
Closes #137.

The output differs from the one requested on the issue due to the underlying library we're using:

```
# kubectl get tenants.capsule.clastix.io -o wide
NAME   NAMESPACE QUOTA   NAMESPACE COUNT   OWNER NAME   OWNER KIND   NODE SELECTOR                  AGE
oil    3                 0                 alice        User         {"kubernetes.io/os":"linux"}   1s
```

### DIFF
--- a/api/v1alpha1/tenant_types.go
+++ b/api/v1alpha1/tenant_types.go
@@ -100,6 +100,7 @@ type TenantStatus struct {
 // +kubebuilder:printcolumn:name="Namespace count",type="integer",JSONPath=".status.size",description="The total amount of Namespaces in use"
 // +kubebuilder:printcolumn:name="Owner name",type="string",JSONPath=".spec.owner.name",description="The assigned Tenant owner"
 // +kubebuilder:printcolumn:name="Owner kind",type="string",JSONPath=".spec.owner.kind",description="The assigned Tenant owner kind"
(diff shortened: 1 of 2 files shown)


bsctl:
> The output differs from the one requested on the issue due to the underlying library we're using:

Is this because of the `apiVersion: apiextensions.k8s.io/v1beta1`?

prometherion:
Nope, Kubernetes CRDs just have certain kind of column types: boolean, date, string, number.

So that's the resulting string conversion of a string/string map.

bsctl (approved):
lgtm

PR #142, opened 2020-11-17
Allowing regex or list of allowed Container Registries per Tenant

### LINKED ISSUE #1: Ensure all images come from a trusted registry declared in the Tenant

prometherion:
Closes #1.

The final spec is the following:

```yaml
kind: Tenant
spec:
  containerRegistries:
    allowed:
    - docker.io
    regex: "internal.registry.\\w.tld"
```

The `containerRegistries` specification key can be set to `null` in order to turn off the feature, making this implementation backward compatible.


prometherion:
What's great here is that we got our first TODO:
https://github.com/clastix/capsule/blob/1fd8e4ad01b14cac4eb671948ead4304765e8f5e/pkg/webhook/registry/validating.go#L34-L35

The actual limitation of Kubebuilder is that we cannot set the scope of webhook, that means they're global: performing dynamic admission control for Pods could lead to a cascading failure in case Capsule is failing, creating a nested loop that can be just solved by deleting the Validating Webhook configuration.

The actual workaround is setting the FailurePolicy to `Ignore` but, in Helm Charts, I would put the _Namespace Selector_ to match the expression `{"key": "capsule.clastix.io/tenant", "operator": "Exists"}` enforcing the failure policy just to play safe with this enforcement.

The same can be applied to other webhooks, just to decrease the pressure against Capsule that would be useless.

bsctl (commented):
We should consider the case where a registry contains a flat structure of images: `registry/image:tag`  instead of `registry/repository/image:tag`.

As main example, having:
```yaml
  containerRegistries:
    allowed:
      - docker.io
    allowedRegex: ''
```

The request to create a pod with image:

```yaml
  containers:
    - name: hello
      image: docker.io/nginx:latest
```

is blocked by:

`admission webhook "pod.capsule.clastix.io" denied the request: Image's container registry docker.io/nginx:latest is forbidden for the current Tenant, one of the following (docker.io)`

The request to create a pod with image:

```yaml
  containers:
    - name: tommy
      image: docker.io/basct/tomcat:latest
```

succeeds.

prometherion:
Are you sure `docker.io` is the resolved registry URL?

I was pretty sure that was `hub.docker.io`: please, can you double check?

bsctl:
> Are you sure `docker.io` is the resolved registry URL?

```
$ docker run -d docker.io/nginx:latest
2cf83cda6065e6ba307c41d8c190e4ecfe7fe6492409d737916999c9748c096e
[node1] (local) root@192.168.0.23 ~
$ docker run -d hub.docker.io/nginx:latest
Unable to find image 'hub.docker.io/nginx:latest' locally
docker: Error response from daemon: Get https://hub.docker.io/v2/: dial tcp: lookup hub.docker.io on 127.0.0.11:53: no such host.
See 'docker run --help'.
[node1] (local) root@192.168.0.23 ~
```

prometherion:
Whoa, this is fuzzy.

Need to think a solution to this, because actually we are _regexing_ to retrieve the registry from un-fully qualified Container image.

I'll be back.

image

prometherion:
ok @bsctl, here we are: latest changes are checking the `map[string]string` provided by the regex, and we fallback to `docker.io` instead of the wrong `hub.docker.io` also for edge cases as `docker.io/nginx:alpine`.

Last but not least, I removed a TODO I was going to add, using the `namespaceSelector` to the Pod's containers image registries and Ingress, PVC, and Network Policy webhooks too: I would say it's an improvement since we're going to receive just the events from Namespaces managed by a Tenant, decreasing the chance of blocking the entire Kubernetes cluster

This last change (webhooks `namespaceSelector` key) must be replicated on Helm Charts too: the `manifest.yaml` doesn't include these changes since I had to use the `kustomize`'s JSON patching feature.

bsctl:
@prometherion may be something went wrong: now no check is done, any registry is allowed.

prometherion:
@bsctl I guess you were using the wrong typed Tenant sample, please try again with the correct one: it has been fixed with last force pushed commit.

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  containerRegistries:
    allowed:
      - docker.io
    allowedRegex: ""
# redacted
```

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: quay.io/srestuff/nginx:alpine
EOF
# Error from server: error when creating "STDIN": admission webhook "pod.capsule.clastix.io" denied the request: Container image quay.io/srestuff/nginx:alpine registry is forbidden for the current Tenant:  use one from the following list (docker.io)
```

bsctl:
@prometherion we miss the case where tag is not provided:

```
$ kubectl -n cap-production run nginx --image=docker.io/nginx 
Error from server: admission webhook "pod.capsule.clastix.io" denied the request:
Container image docker.io/nginx registry is forbidden for the current Tenant: use one from the following list (docker.io, quay.io)

$ kubectl -n cap-production run nginx --image=docker.io/nginx:latest
pod/nginx created

$ kubectl -n cap-production run nginx --image=quay.io/redhattraining/hello-world-nginx
Error from server: admission webhook "pod.capsule.clastix.io" denied the request:
Container image quay.io/redhattraining/hello-world-nginx registry is forbidden for the current Tenant: use one from the following list (docker.io, quay.io)

$ kubectl -n cap-production run nginx --image=quay.io/redhattraining/hello-world-nginx:latest
pod/hello created
```

prometherion:
@bsctl had I told you how are you superiorly good at QA? 🥇 

Thinking to use the following regex, rather: `(((?P<registry>[a-zA-Z0-9-.]+)\/)?((?P<repository>[a-zA-Z0-9-.]+)\/))?(?P<image>[a-zA-Z0-9-.]+)(:(?P<tag>[a-zA-Z0-9-.]+))?`

bsctl:
@prometherion can we add these test cases to our CI pipeline?

prometherion:
> @prometherion can we add these test cases to our CI pipeline?

Absolutely yes, actually we don't need an E2E but easily the [unit test case](https://github.com/clastix/capsule/blob/eaf9a822272785245854ab2c7c5dc122699622a5/api/v1alpha1/domain/registry_test.go#L33-L62), there.

prometherion:
@bsctl I provided the required changes along with the test case
https://github.com/clastix/capsule/blob/5f2047b2f0faba811a71433133afcbd974fb28bf/api/v1alpha1/domain/registry_test.go#L63-L68

Should be ok: another QA run, please?

bsctl (approved):

PR #147, opened 2020-11-30
Helm Charts are now inside of the repository

prometherion:
Should close clastix/capsule-helm-chart#32 but using a new repository hosting as Helm Chart one.


prometherion:
@MaxFedotov honestly, since we're incorporating Helm Chart in the repository, I'd go set up the CI installation through Helm rather than `kustomize` so we could save a lot of resources rather than downloading tons of Go modules.

WDYT? /cc @bsctl

bsctl:
@prometherion is this only for CI? Users still have the two options, i.e `kustomize` and `helm`?

prometherion:
I'd say `kustomize` is for developers and still required to generate manifests that will feed Helm.

End users are expected to use only Helm, from source or the repository hosted on https://clastix.github.io/charts

bsctl:
I would leave an alternative to Helm for end users. Many people hate Helm :(
Should we consider also to use a naked single `capsule-operator.yaml` file so that the end user just download, eg. with `wget` or `curl` and then `kubectl apply`?

prometherion:
Honestly, Helm became the official way to install apps on Kubernetes, despite a haggard minority ranting about it.

It's has been graduated CNCF and we already provide manifests can be easily used through `kustomize` that is shipped along with `kubectl`.

Said so, we cannot focus on this: we already provide two options, end users will pick the preferred one.

bsctl:
> Said so, we cannot focus on this: we already provide two options, end users will pick the preferred one.

From previous comment, my understanding was Helm will be the only option. Got it now.
+1

bsctl (approved):
LGTM

PR #153, opened 2020-12-05
Only owner Tenant specification key is mandatory

### LINKED ISSUE #145: Changing required fields for Tenant resource
I would review the required fields in Tenant resource. Currently (v0.0.3), in `capsule.clastix.io/v1alpha1`, we have:

|field                      |required   |webhook    |controller |
|---------------------------|-----------|-----------|-----------|
|owner                      |*          |*          |           |
|nodeSelector               |           |           |*          |
|namespaceQuota             |*          |*          |           |
|namespacesMetadata         |           |           |*          |
|servicesMetadata           |           |           |*          |
|ingressClasses             |*          |*          |           |
|storageClasses             |*          |*          |           |
|containerRegistries        |           |*          |           |
|additionalRoleBindings     |           |           |*          |
|resourceQuotas             |           |           |*          |
|limitRanges                |*          |           |*          |
|networkPolicies            |           |*          |           |

To make Capsule more flexible and address more use cases, we can leave only the `owner` field as mandatory. All the other fields will be optional.

@prometherion @MaxFedotov @gdurifw what do you think? What prevents this from technical pov?

prometherion:
Closes #145, letting only `.spec.owner` be required and making all the others optional.

As a side effect, the e2e cases needed to be refactored a little bit since we can avoid declaring empty values thanks to the struct tag `omitempty`.

### DIFF
--- a/api/v1alpha1/tenant_func.go
+++ b/api/v1alpha1/tenant_func.go
@@ -23,7 +23,11 @@ import (
 )
 
 func (t *Tenant) IsFull() bool {
-	return t.Status.Namespaces.Len() >= int(t.Spec.NamespaceQuota)
+	// we don't have limits on assigned Namespaces
+	if t.Spec.NamespaceQuota == nil {
+		return false
+	}
+	return t.Status.Namespaces.Len() >= int(*t.Spec.NamespaceQuota)
 }
 
 func (t *Tenant) AssignNamespaces(namespaces []corev1.Namespace) {
--- a/api/v1alpha1/tenant_types.go
+++ b/api/v1alpha1/tenant_types.go
@@ -23,55 +23,42 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// +kubebuilder:validation:Minimum=1
-type NamespaceQuota uint
-
 type AdditionalMetadata struct {
-	// +nullable
-	AdditionalLabels map[string]string `json:"additionalLabels"`
-	// +nullable
-	AdditionalAnnotations map[string]string `json:"additionalAnnotations"`
+	AdditionalLabels      map[string]string `json:"additionalLabels,omitempty"`
+	AdditionalAnnotations map[string]string `json:"additionalAnnotations,omitempty"`
 }
 
 type StorageClassesSpec struct {
-	// +nullable
-	Allowed StorageClassList `json:"allowed"`
-	// +nullable
-	AllowedRegex string `json:"allowedRegex"`
+	Allowed      StorageClassList `json:"allowed,omitempty"`
+	AllowedRegex string           `json:"allowedRegex,omitempty"`
 }
 
 type IngressClassesSpec struct {
-	// +nullable
-	Allowed IngressClassList `json:"allowed"`
-	// +nullable
-	AllowedRegex string `json:"allowedRegex"`
+	Allowed      IngressClassList `json:"allowed,omitempty"`
+	AllowedRegex string           `json:"allowedRegex,omitempty"`
 }
 
 type ContainerRegistriesSpec struct {
-	// +nullable
-	Allowed RegistryList `json:"allowed"`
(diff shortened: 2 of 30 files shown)


bsctl (commented):
@prometherion testing features addition. I found that a adding `nodeSelector: pool=cmp` to a tenant already having namespaces, the annotation `scheduler.alpha.kubernetes.io/node-selector: pool=cmp` is correctly applied to all namespaces (including namespaces already provisioned). However, when removing the `nodeSelector: pool=cmp` from the tenant, the annotation is still there. Is this behaviour as expected?

--
It looks this behavior is in common with other features, for example `namespacesMetadata`

bsctl:
@prometherion 
Storage and Ingress Classes are still required.
I have a tenant with only owner

```
kind: Tenant
metadata:
  name: tenant
spec:
  owner:
    kind: User
    name: alice
status:
  namespaces:
  - sample-development
  - sample-investigation
  - sample-production
```

- Creating an Ingress without Ingress Class, we get the error: `admission webhook "ingress-v1beta1.capsule.clastix.io" denied the request: A valid Ingress Class must be used`

- Creating an Ingress with Ingress Class `example`, we get the error: `admission webhook "ingress-v1beta1.capsule.clastix.io" denied the request: Ingress Class example is forbidden for the current Tenant`

- Creating a PVC without Storage Class, we get the error: `admission webhook "pvc.capsule.clastix.io" denied the request: A valid Storage Class must be used`

- Creating a PVC with a Storage Class `example`, we get the error: `admission webhook "pvc.capsule.clastix.io" denied the request: Storage Class example is forbidden for the current Tenant`

MaxFedotov:
@bsctl I think we come to issue that we discussed about a month ago in a slack (https://clastix.slack.com/archives/C017HQ6LTEE/p1599827942009000) about allow-first or deny-first policies

bsctl:
@MaxFedotov @prometherion 
I remember we reached consensus on the following:

1. if no ingress/storage are specified in the tenant spec -> we assume the cluster admin does not want to enforce that tenant about ingress/storage classes -> all ingress/storage classes are permitted

2. if some ingress/storage classes (both listed and regex) are specified in the tenant spec -> we assume the cluster admin does want to enforce that tenant about ingress/storage classes -> only specified ingress/storage classes (both listed and regex) are permitted

The user checks the annotations on the namespaces:

- if there are annotations, the user knows he has to use only the listed classes
- if there are no annotations, the user knows he can use any classes or no class at all.
 
This PR should cover both the cases above.

Hope I'm not missing something.

prometherion:
Ok, we have to use a nil pointer on the Ingress and Storage classes, I'd say.

prometherion:
@bsctl please, take a look at this: should be ready to get it merged.

bsctl (commented):
The behaviour described [here](https://github.com/clastix/capsule/pull/153#issuecomment-741693521) is no more there and now it is as [expected](https://github.com/clastix/capsule/pull/153#issuecomment-741739725).

bsctl (commented):
@prometherion wonder to know your PoV on [this](https://github.com/clastix/capsule/pull/153#pullrequestreview-546460284). Maybe it is not related to this PR.

prometherion:
> @prometherion wonder to know your PoV on [this](https://github.com/clastix/capsule/pull/153#pullrequestreview-546460284). Maybe it is not related to this PR but it looks a general behaviour. Should we try to fix it?

Completely lost the focus on this, I'll investigate it!

prometherion:
> @prometherion wonder to know your PoV on [this](https://github.com/clastix/capsule/pull/153#pullrequestreview-546460284). Maybe it is not related to this PR but it looks a general behaviour. Should we try to fix it?

You were right, noticed a bug and fixed with b7484ec: in case of nil value, we weren't deleting the key, keeping it there until a new value was set.

bsctl:
@prometherion just few quick tests. LGTM

PR #156, opened 2020-12-06
Enhancing Helm Chart lifecycle

### LINKED ISSUE #155: On Helm upgrade, the CA/TLS secrets are overwritten with empty data
# Bug description

This is happening upon Helm upgrade, forcing a new CA/TLS generation causing an outage in case of an upgrade due to a new release.

# How to reproduce

1. `helm upgrade --install capsule ./charts/capsule -n capsule-system --create-namespace`
2. Wait for get it up and running
3. Once done, check the Secrets in `capsule-system`, should contain 2 keys
4. `helm upgrade --install capsule ./charts/capsule -n capsule-system --create-namespace`
5. The Secrets in `capsule-system` have been erased, no keys in there

# Expected behavior

Upon `helm upgrade` the Secret resources should not be overwritten.

# Logs

N.R.

# Additional context

- Capsule version: `0.0.3`
- Helm Chart version: `0.0.6`
- Kubernetes version: `N.R.`

prometherion:
Closes #155.

Secret resources are going to be created only at installation and removed using the pre-delete hook thas has been improved as the post-install one.

### DIFF
--- a/charts/capsule/Chart.yaml
+++ b/charts/capsule/Chart.yaml
@@ -21,7 +21,7 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.0.6
+version: 0.0.7
 
 # This is the version number of the application being deployed.
 # This version number should be incremented each time you make changes to the application.
--- a/charts/capsule/templates/_helpers.tpl
+++ b/charts/capsule/templates/_helpers.tpl
@@ -67,3 +67,24 @@ Create the fully-qualified Docker image to use
 {{- define "capsule.fullyQualifiedDockerImage" -}}
 {{- printf "%s:%s" .Values.manager.image.repository ( .Values.manager.image.tag | default (printf "v%s" .Chart.AppVersion) ) -}}
 {{- end }}
+
+{{/*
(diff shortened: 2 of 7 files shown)


MaxFedotov:
@prometherion Everything looks good. The only thing I'm worried about is moving helmHook commands to values.yaml file. I don't think that they should be modified by user, so maybe it's better to move them into helpers?

prometherion:
Thanks for the feedback, @MaxFedotov.

The reason behind putting the command into the values file was to make it more readable, and it's not intended to be modified by the end-user.

If it doesn't make sense, I'll put them back into the Helm template.

MaxFedotov:
> The reason behind putting the command into the values file was to make it more readable, and it's not intended to be modified by the end-user.
Yeah, I understand :) But for me, values are something, that I can tune for my env when I'm running deployment. How do you think, will it be possible to move this into _helpers file?

prometherion:
Defo, going to put into Helpers.

prometherion:
@MaxFedotov please, review squashed commit: rather than Helpers, since this is going to used just into the template, defined as variable there: WDYT?

bsctl (approved):
@prometherion followed steps in #155. It looks LGTM.

PR #161, opened 2020-12-10
Enforcing Service external IPs

### LINKED ISSUE #159: Policy to restrict External IPs in Services [CVE-2020-8554]
# Describe the feature
We need to add a policy to restrict External IPs in services in order to mitigate CVE-2020-8554. Ref. [here](https://github.com/kubernetes/kubernetes/issues/97076) and [here](https://github.com/kubernetes/kubernetes/issues/97110)

1. CVE-2020-8554 affects multi-tenancy environments
2. The cluster admin creates a tenant specifying the allowed External IP CIDRs 

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  owner:
    kind: User
    name: alice
  externalServiceIPs:
     allowed:
     - 10.0.0.0/8
     - 192.168.0.0/16
    
```

3. This is going to be attached to the Tenant
4. The tenant owner creates a Service by specifying the External IPs: only listed IPs are allowed 
5. If `externalServiceIPs` is not specified, then no External IPs in Services are allowed 

# Expected behavior
The cluster admin can mitigate CVE-2020-8554

prometherion:
Closes #159.


bsctl (commented):
@prometherion with the current implementation, omitting `externalServiceIPs` from tenant manifest, the admission control is not in place and all IPs are permitted.

It seems we have following options to block all IPs:

1. if the `externalServiceIPs` spec is not declared in the tenant, then no IPs are permitted at all, as stated in #159. This will be an exception in the general approach where if a spec is not declared, the related admission controller is not applied, eg. `IngressClasses`. Actually, this different approach could be a bit confusing for the end user.

2. permit an empty value to the allowed list of IPS, for example

```yaml
kind: Tenant
spec:
  externalServiceIPs:
    allowed: []
```

Option 2 actually works. However, since `externalServiceIPs` is not mandatory, it is responsibility of the cluster admin to apply that spec in order to mitigate CVE-2020-8554.

Wonder to know your point of view. Thanks.

prometherion:
> omitting externalServiceIPs from tenant manifest, the admission control is not in place and all IPs are permitted.

This looks good to me, it's an expected behavior according to the current implementation and it reflects what we achieved with Storage and Ingress classes.

If CVE must be mitigated, a valid `.spec.externalServiceIPs.allowed` set of CIDR blocks has to be put in place.

bsctl:
LGTM

PR #162, opened 2020-12-11
Issues/157 hostnames

### LINKED ISSUE #157: Check the hostname that an Ingress can use
# Describe the feature
It would be nice to have additional admission controls on the Ingresses resources created by the tenant owners:

 A. white list of admitted hostnames for a given tenant.

For example:

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  ingressHostnames:
     allowed:
     - my.oil.acmecorp.com
     - my.svc.acme.com
     allowedRegex: "^.*$"
```

The above is useful to assign domain(s) to the tenant.

B. check of unicity on the hostname, i.e. deny when `my.oil.acmecorp.com` is already taken by another Ingress (same or different tenant)

1. Cluster admin set the allowed hostnames per tenant
2. Tenant owner creates/updates a new ingress with given hostnames
3. Capsule admission control checks the validity of the hostnames and denies in case of not allowed (case A)
4. Capsule admission control checks the unicity of the hostnames and denies in case of duplicated (case B)

# Expected behavior
Check the hostname that an Ingress can use

paolocarta:
First draft for validating the ingress' hostnames. My first PR in Go, I hope is not too bad!

At the moment I do not perform yet any validation at the tenant level to ensure there are no duplicate domains.

I need to add some tests, I need a bit of study how to do that in Go for such a project.

close #157


pkg/webhook/ingress/validating.go
```diff
@@ -56,74 +59,80 @@ func (w *webhook) GetPath() string {
...
 }
 
-type handler struct{}
+type handler struct {
+	Log logr.Logger
+}
 
 func Handler() capsulewebhook.Handler {
-	return &handler{}
+	return &handler{Log: ctrl.Log.WithName("controllers").WithName("Tenant")}
```
paolocarta, line 67: Adding a logger for logging purposes during debugging.

pkg/webhook/ingress/validating.go
```diff
@@ -155,6 +164,35 @@ func (r *handler) validateIngress(ctx context.Context, c client.Client, object I
 		return admission.Errored(http.StatusBadRequest, NewIngressClassForbidden(*ingressClass, *tnt.Spec.IngressClasses))
 	}
 
+	//TODO extract logic below into a method
```
paolocarta, line 167: Maybe I should extract the logic into a method, since it's getting bigger. What do you think?

pkg/webhook/ingress/validating.go
```diff
@@ -155,6 +164,35 @@ func (r *handler) validateIngress(ctx context.Context, c client.Client, object I
...
+	if tnt.Spec.IngressHostnames == nil {
+		return admission.Allowed("")
+	}
+
+	valid = false
+	matched = false
+	hostnames := ingress.Hostnames()
+	if len(hostnames) > 0 {
+		valid = tnt.Spec.IngressHostnames.Allowed.AreStringsInList(hostnames)
```
paolocarta, line 176: Here I verify if all strings are in the allowed list.

pkg/webhook/ingress/validating.go
```diff
@@ -155,6 +164,35 @@ func (r *handler) validateIngress(ctx context.Context, c client.Client, object I
...
+	}
+
+	allowedRegex := tnt.Spec.IngressHostnames.AllowedRegex
+	if len(allowedRegex) > 0 {
+		matched = allowedRegex.MatchesAllStrings(hostnames)
```
paolocarta, line 181: I must verify all hostnames in the ingress match the regex.

paolocarta:
Do I still need to adapt anything with the indexer?

.gitignore
```diff
@@ -24,3 +24,5 @@ bin
 *~
 
 hack/*.kubeconfig
+
+alice-oil.kubeconfig
```
prometherion, line 28: hum, why `alice-oil.kubeconfig` here? It should be generated inside the `hack` folder, and so ignored.

prometherion, line 28: hum, why `alice-oil.kubeconfig` here? It should be generated inside the `hack` folder, and so ignored.

api/v1alpha1/tenant_types.go
```diff
@@ -38,6 +38,13 @@ type IngressClassesSpec struct {
 	AllowedRegex string           `json:"allowedRegex,omitempty"`
 }
 
+type IngressHostnamesSpec struct {
+	// +nullable
```
prometherion, line 42: let's remove the `nullable` marker here: we don't need it if we mark the `IngressHostnamesSpec` as `omitempty`: refer to this.

https://github.com/clastix/capsule/blob/98e441f1e9a9e24630624fcb8c68ff5039b59e28/api/v1alpha1/tenant_types.go#L61-L63

https://github.com/clastix/capsule/blob/98e441f1e9a9e24630624fcb8c68ff5039b59e28/api/v1alpha1/tenant_types.go#L31-L44

pkg/webhook/ingress/types.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package ingress
 
 import (
+	"github.com/go-logr/logr"
```
prometherion, line 20: please, sort imports:

1. standard library
2. external packages
3. internal ones (`github.com/clastix/capsule`)

pkg/webhook/ingress/validating.go
```diff
@@ -21,16 +21,19 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	ctrl "sigs.k8s.io/controller-runtime"
```
prometherion, line 24: Imports sorting required.

pkg/webhook/ingress/validating.go
prometherion: No problem, let's remove it once ready to get it merged: there's a better way to debug webhooks, although pretty difficult to setup.

https://github.com/clastix/capsule/blob/master/contributing.md#run-capsule-locally

pkg/webhook/ingress/validating.go
```diff
@@ -56,74 +59,80 @@ func (w *webhook) GetPath() string {
...
 			return admission.Errored(http.StatusBadRequest, err)
 		}
 
-		return r.validateIngress(ctx, client, i)
+		return r.validateIngress(ctx, client, ingress)
 	}
 }
 
 func (r *handler) OnDelete(client client.Client, decoder *admission.Decoder) capsulewebhook.Func {
+
```
prometherion, line 95: I saw this new line at the beginning of every function in the files you edited: is this an IDE linting option?

pkg/webhook/ingress/validating.go
prometherion: Please, refer to the right name: Go doesn't provide __method__ but rather __struct functions__ :)

But I'd say webhook is quite atomic enough, no need to split into micro functions unless they're shared, but I'd say this is not the case.

pkg/webhook/ingress/validating.go
prometherion: LGTM, although  a different approach to what we implemented with other components such as the Container Registry check

https://github.com/clastix/capsule/blob/98e441f1e9a9e24630624fcb8c68ff5039b59e28/pkg/webhook/registry/validating.go#L85-L94

It's just a smell of code stuff, maybe iterating over the array would make it easier to understand the business logic behind the webhook.

Let's start the discussion.

pkg/webhook/router.go
```diff
@@ -34,11 +35,11 @@ func Register(mgr controllerruntime.Manager, webhookList ...Webhook) error {
 		return nil
 	}
 
-	s := mgr.GetWebhookServer()
-	for _, wh := range webhookList {
-		s.Register(wh.GetPath(), &webhook.Admission{
+	whServer := mgr.GetWebhookServer()
```
prometherion, line 38: please, don't rename variables: it's not a matter of the PR.

prometherion (changes requested):
Looks good, thanks @paolocarta! Some minor request changes and a little doubt about the search in strings we have to perform: looking forward to your point of view to continue the great work accomplished!

pkg/webhook/ingress/validating.go
paolocarta: My idea here would be to have a general purpose matching between two arrays. I need to verify that all elements in the "hostnames" array are contained in the allowed list.

.gitignore
paolocarta: In the script there is:
`cat > ${USER}-${TENANT}.kubeconfig`
If you execute it from the main folder with:
`./hack/create-user.sh alice oil`
You get the file in the main folder.

In case I could adapt it in a additional PR. 

WDYT?

api/v1alpha1/tenant_types.go
paolocarta: done.

pkg/webhook/ingress/types.go
paolocarta: It should be after the imports of `k8s.io` then? There are no standard go libraries in this file.

pkg/webhook/ingress/validating.go
paolocarta: done.

pkg/webhook/ingress/validating.go
paolocarta: Yes, just a personal preference. I can remove it.

pkg/webhook/router.go
paolocarta: Not in the scope you are right.

Just a general remark, don't you think that names like "wh" slow down understanding of the code? Maybe I am obsessed with uncle bob's clean code! :-D

.gitignore
prometherion: Oh, it makes sense: let's __gitignore__ `*.kubeconfig` everywhere.

```
sed -ie 's#hack/*.kubeconfig#*.kubeconfig#'
```

pkg/webhook/ingress/validating.go
```diff
@@ -21,16 +21,19 @@ import (
...
 	networkingv1 "k8s.io/api/networking/v1"
 	networkingv1beta1 "k8s.io/api/networking/v1beta1"
 	"k8s.io/apimachinery/pkg/fields"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/clastix/capsule/api/v1alpha1"
 	capsulewebhook "github.com/clastix/capsule/pkg/webhook"
+	"github.com/go-logr/logr"
```
prometherion: My bad, I was pointing to this @paolocarta

pkg/webhook/ingress/types.go
prometherion: My bad, this snippet is good: I was rather pointing to [this](https://github.com/clastix/capsule/pull/162/files/47a20ab0f0705f14d8a0144355a5b40148ba70b3#r542397019)

pkg/webhook/ingress/validating.go
prometherion: You're not the first, indeed! :)

We're not documenting this in the coding style documented for Capsule, but we can create an issue to document this, too.

Please, let's keep the current style.

pkg/webhook/router.go
prometherion: Uncle Bob is always right, mine is a heritage from other Go codebases where they used to assign really short variables, as you can notice in `mgr` that is easy to understand as `manager`.

I thought the same for `wh` (`webhook`) and `s` as `server` but seems I failed (__but at what cost?__).

Definitely, we have also to take care of the developer experience in order to onboard as fast as possible other people, so I'd raise an Issue to discuss this in-depth since inclusivity can start also from variables naming!

.gitignore
paolocarta: done.

pkg/webhook/ingress/validating.go
prometherion: I thought about this and checking a list of strings against doesn't return the missed intersection, or rather the set of strings not in the defined list.

Since in the end we're iterating over using `range`, let's avoid to use a func and implement this loop here so we can gather the item not in the array.

pkg/webhook/router.go
```diff
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+
```
prometherion, line 24: No need of new line, imports not sorted correctly.

pkg/webhook/ingress/validating.go
paolocarta: fine for me.

config/crd/bases/capsule.clastix.io_tenants.yaml
```diff
@@ -134,6 +134,18 @@ spec:
                 allowedRegex:
                   type: string
               type: object
+            ingressHostnames:
```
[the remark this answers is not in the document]
paolocarta: done.

pkg/webhook/router.go
paolocarta: done.

bsctl:
@paolocarta @prometherion just a small question here.

What about to use 

```
            ingressHostNames:
              properties:
```

instead of

```
            ingressHostnames:
              properties:
```

in CRD?

prometherion:
I'd say the context is the following: https://en.m.wikipedia.org/wiki/Hostname

So camel casing the term itself would be bad, since hostname is the term.

api/v1alpha1/ingress_hostnames_list.go
```diff
@@ -0,0 +1,43 @@
...
+func (hostnames IngressHostnamesList) Len() int {
+	return len(hostnames)
+}
+
+func (hostnames IngressHostnamesList) Swap(i, j int) {
+	hostnames[i], hostnames[j] = hostnames[j], hostnames[i]
+}
+
+func (hostnames IngressHostnamesList) Less(i, j int) bool {
+	return strings.ToLower(hostnames[i]) < strings.ToLower(hostnames[j])
```
prometherion, line 35: We noticed an odd behavior reported in #177, so better to remove the `strings.ToLower` otherwise the binary search would be broken.

e2e/selecting_tenant_fail_test.go
```diff
@@ -77,27 +77,27 @@ var _ = Describe("creating a Namespace without a Tenant selector when user owns
 	It("should fail", func() {
 		ns := NewNamespace("fail-ns")
 		By("user owns 2 tenants", func() {
-			EventuallyCreation(func() error { return k8sClient.Create(context.TODO(), t1)}).Should(Succeed())
-			EventuallyCreation(func() error { return k8sClient.Create(context.TODO(), t2)}).Should(Succeed())
+			EventuallyCreation(func() error { return k8sClient.Create(context.TODO(), t1) }).Should(Succeed())
```
prometherion, line 80: Please, remove additional and unrequired spaces in this file, unless for a specific reason.

prometherion (commented):
Besides minor things, I'd say we're in a good shape!

@paolocarta what are we still missing to get this merged?

Let's ignore tests, I can deal with it on my own since BDD is quite hard.

api/v1alpha1/ingress_hostnames_list.go
paolocarta: done.

e2e/selecting_tenant_fail_test.go
paolocarta: done

prometherion:
@paolocarta JFI I provided minor improvement and an e2e feature test, let's see if the test suite is green so we can merge this! 🎉

PR #169, opened 2020-12-20
Adding linters and aligning code

### LINKED ISSUE #166: Define required golangci linters
Actually, we're just running the default ones, we should consider switching on more strict linters.

prometherion:
Closes #166.


prometherion:
These changes could affect the work in progress by @PaoloCarta: thanks to the additional linters I discovered the (obvious) duplication of code for the StorageClass, IngressClass, and Registry allowed lists with exact and regex matching.

So I decided to perform refactoring of this to avoid the duplication of code, and the result seems neat to me but I could expect some pain performing a rebase of your work on #162: so, if this is going to happen, we can delay this PR after yours.

Just let me know!

PR #172, opened 2020-12-20
Eventually refactoring and minor fixes

### LINKED ISSUE #120: Refactoring e2e test suite

prometherion:
Closes #120.

### DIFF
--- a/e2e/container_registry_test.go
+++ b/e2e/container_registry_test.go
@@ -33,7 +33,7 @@ import (
 var _ = Describe("enforcing a Container Registry", func() {
 	tnt := &v1alpha1.Tenant{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "additional-role-binding",
+			Name: "container-registry",
 		},
 		Spec: v1alpha1.TenantSpec{
 			Owner: v1alpha1.OwnerSpec{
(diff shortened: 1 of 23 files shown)


e2e/container_registry_test.go
```diff
@@ -33,7 +33,7 @@ import (
 var _ = Describe("enforcing a Container Registry", func() {
 	tnt := &v1alpha1.Tenant{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "additional-role-binding",
+			Name: "container-registry",
```
prometherion, line 36: Just a typo in the Tenant name, not so critical.

e2e/container_registry_test.go
```diff
@@ -44,51 +44,35 @@ var _ = Describe("enforcing a Container Registry", func() {
...
 	It("should add labels to Namespace", func() {
 		ns := NewNamespace("registry-labels")
 		NamespaceCreation(ns, tnt, defaultTimeoutInterval).Should(Succeed())
-
-		Eventually(func() bool {
-			if err := k8sClient.Get(context.Background(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
-				return false
+		Eventually(func() (ok bool) {
+			Expect(k8sClient.Get(context.Background(), types.NamespacedName{Name: ns.Name}, ns)).Should(Succeed())
+			ok, _ = HaveKeyWithValue("capsule.clastix.io/allowed-registries", "docker.io,docker.tld").Match(ns.Annotations)
```
prometherion, line 64: Although is an upside-down assertion, this works well since we don't know when the Namespace has been totally processed by the controller.

e2e/resource_quota_exceeded_test.go
```diff
@@ -166,63 +165,41 @@ var _ = Describe("exceeding Tenant resource quota", func() {
 				}
 				_, err = cs.AppsV1().Deployments(namespace).Create(context.TODO(), d, metav1.CreateOptions{})
 				return
-			}, 15*time.Second, time.Second).Should(Succeed())
```
prometherion, line 169: This is legacy hard-coded time values.

e2e/resource_quota_exceeded_test.go
```diff
@@ -166,63 +165,41 @@ var _ = Describe("exceeding Tenant resource quota", func() {
...
 			By("ensuring the status has been blocked with actual usage", func() {
 				Eventually(func() corev1.ResourceList {
 					_ = k8sClient.Get(context.TODO(), types.NamespacedName{Name: n, Namespace: ns}, rq)
 					return rq.Status.Hard
-				}, 15*time.Second, time.Second).Should(Equal(rq.Status.Used))
+				}, defaultTimeoutInterval, defaultPollInterval).Should(Equal(rq.Status.Used))
 			})
 			By("creating an exceeded Pod", func() {
-				d := &v1.Deployment{
+				pod := &corev1.Pod{
```
prometherion, line 185: Rather than creating Deployment, checking ReplicaSet and finally conditions, is it more suitable to run a single Pod!

e2e/service_metadata_test.go
```diff
@@ -51,49 +54,40 @@ var _ = Describe("creating a Service/Endpoint/EndpointSlice for a Tenant with ad
 					"clastix.io/custom-annotation": "buzz",
 				},
 			},
-		},
-	}
-	epsCR := &rbacv1.ClusterRole{
```
prometherion, line 56: Taking advantage of the additional role bindings offered with latest versions of Capsule, using the `endpointslice-controller`.

PR #188, opened 2021-01-14
Non preserving CRD unknown fields

### LINKED ISSUE #187: Tenant CRD should not have preserveUnknownFields: true
The CRD attribute `preserveUnknownFields` is allowing to put non structured data into the Tenant definition.

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  ingressClasses:
    allowed:
      - caas
    allowedRegex: ''
  mealsOfTheDay:
    allowed:
      - pasta
      - pizza
      - spaghetti
  owner:
    kind: User
    name: alice
```

By default, the `.spec.preserveUnknownFields` field is set to `true`: it's not critical but would be nice to be informed of typos during the specification apply.

prometherion:
Closes #187.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/clastix/capsule:$(VERSION)
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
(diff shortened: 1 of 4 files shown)


bsctl (approved):
lgtm

`error: error validating "/private/var/folders/07/p2qbsxp97db1ht5sbcr83xc80000gn/T/a695949ec576901ac293efd084fce249/resource.yaml": error validating data: ValidationError(Tenant.spec): unknown field "namespaceQuotas" in io.clastix.capsule.v1alpha1.Tenant.spec; if you choose to ignore these errors, turn validation off with --validate=false`

PR #209, opened 2021-03-01
[Helm] Make Jobs containers image configurable

unai-ttxu:

### DIFF
--- a/charts/capsule/templates/deployment.yaml
+++ b/charts/capsule/templates/deployment.yaml
@@ -81,7 +81,7 @@ spec:
             allowPrivilegeEscalation: false
         - name: kube-rbac-proxy
           image: {{ .Values.proxy.image.repository }}:{{ .Values.proxy.image.tag }}
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: {{ .Values.proxy.image.pullPolicy }}
           args:
           - --secure-listen-address=0.0.0.0:8443
           - --upstream=http://127.0.0.1:8080/
--- a/charts/capsule/templates/post-install-job.yaml
+++ b/charts/capsule/templates/post-install-job.yaml
@@ -28,7 +28,8 @@ spec:
       restartPolicy: Never
       containers:
       - name: post-install-job
-        image: "bitnami/kubectl:1.18"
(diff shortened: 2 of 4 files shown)


charts/capsule/templates/deployment.yaml
```diff
@@ -81,7 +81,7 @@ spec:
             allowPrivilegeEscalation: false
         - name: kube-rbac-proxy
           image: {{ .Values.proxy.image.repository }}:{{ .Values.proxy.image.tag }}
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: {{ .Values.proxy.image.pullPolicy }}
```
prometherion, line 84: What about using a template function as we did with the manager container?

https://github.com/clastix/capsule/blob/4f34483dee32bcd8ff88c7590f1ab8e5cdcdfa9e/charts/capsule/templates/_helpers.tpl#L64:L69

prometherion (changes requested):
Thanks for this contribution, really appreciated!

Just a small request change, otherwise LGTM! 💪🏻

charts/capsule/templates/post-install-job.yaml
```diff
@@ -28,7 +28,8 @@ spec:
       restartPolicy: Never
       containers:
       - name: post-install-job
-        image: "bitnami/kubectl:1.18"
+        image: {{ .Values.jobs.image.repository }}:{{ .Values.jobs.image.tag }}
```
prometherion, line 31: What about using a template function as we did with the manager container?

https://github.com/clastix/capsule/blob/4f34483dee32bcd8ff88c7590f1ab8e5cdcdfa9e/charts/capsule/templates/_helpers.tpl#L64:L69

unai-ttxu: Great!

I'll adapt the `kube-rbac-proxy` as well in order to make manage all images the same way.

unai-ttxu:
Hi @prometherion,

Thank you so much for the feedback.

I've included the requested changes and ensure the same image managament for all components.

Let me know if it looks good to you.

Thanks!

prometherion (approved):

Issue #207, opened 2021-03-01
Allow by default multiple tenants reserve the same allowed ingress domain


bsctl:
# Describe the feature
Multiple tenants cannot be allowed to use same ingress domain as reported in #206.

By introducing a new command switch as in #206 the default value `false` leads to the same behaviour. It would be nice to have the switch set to `true` by default since this is the most common scenario. However, we decided to set the default to `true` in a new major release v0.1.0 since doing it in the current v0.0.5 represents a breaking change for current users.

# What would the new user story look like?

A new command switch in Capsule to disable the check on allowed ingress domains:

`--allow-multiple-ingress-domains=true` (default value)

# Expected behavior
Having the switch above set to `true` it is possible to have multiple tenants with same allowed ingress domains

```yaml
apiVersion: capsule.clastix.io/v1alpha1
kind: Tenant
metadata:
  name: oil
spec:
  owner:
    kind: User
    name: alice
  ingressHostnames:
     allowed:
     - *.federation.net
```

```yaml
apiVersion: capsule.clastix.io/v1alpha1
kind: Tenant
metadata:
  name: gas
spec:
  owner:
    kind: User
    name: alice
  ingressHostnames:
     allowed:
     - *.federation.net
```

Each of the two tenants above can define an ingress having the same domain `*.federation.net`. For example: `webapp.federation.net` in tenant `oil` and `myapp.federation.net` in tenet `gas`.

prometherion:
@bsctl I think this is the root of #358, would be better to have a single issue to track: WDYT closing this in favor of the latter?

bsctl:
Closing in favor of #358

PR #214, opened 2021-03-04
Allowing multiple Tenant resources to share same exact list of allowed hostnames

### LINKED ISSUE #206: Allow multiple tenants reserve the same allowed ingress domain
# Describe the feature
Currently it not possible to create multiple tenants allowing the same ingress domain as introduced by #162. There are cases, where this is a too strict limitation as reported (incorrectly) by #205.  It would be nice to have a command switch in Capsule to disable the check on allowed ingress domains 

A new  command switch in Capsule to disable the check on allowed ingress hostnames:

`--allow-multiple-ingress-domains=false` (default value)

# Expected behavior
Having the switch above set to `true` it is possible to have multiple tenants with same allowed ingress domains

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  owner:
    kind: User
    name: alice
  ingressHostnames:
     allowed:
     - *.federation.net
```

```yaml
kind: Tenant
metadata:
  name: gas
spec:
  owner:
    kind: User
    name: alice
  ingressHostnames:
     allowed:
     - *.federation.net
```

Each of the two tenants above can define an ingress having the same domain `*.federation.net`. For example: `webapp.federation.net` in tenant `oil` and `myapp.federation.net` in tenet `gas`.

prometherion:
Closes #206.

Editing the original message due to some mess during PRs.

A new CLI flag (`--allow-tenant-ingress-hostnames-collision`) is added, defaulted to `false` to preserve backward compatibility: in case of toggling, Capsule will skip the Tenant validation that checks for hostnames collision in the key `Spec.ingressHostnames.allowed`.

I summon @bsctl: we need your super QA powers, here!

### DIFF
--- a/charts/capsule/Chart.yaml
+++ b/charts/capsule/Chart.yaml
@@ -21,7 +21,7 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.0.16
+version: 0.0.17
 
 # This is the version number of the application being deployed.
 # This version number should be incremented each time you make changes to the application.
--- a/charts/capsule/README.md
+++ b/charts/capsule/README.md
@@ -65,6 +65,7 @@ Parameter | Description | Default
 `manager.options.capsuleUserGroup` | Override the Capsule user group | `capsule.clastix.io`
 `manager.options.protectedNamespaceRegex` | If specified, disallows creation of namespaces matching the passed regexp | `null`
 `manager.options.allowIngressHostnameCollision` | Allow the Ingress hostname collision at Ingress resource level across all the Tenants | `true`
+`manager.options.allowTenantIngressHostnamesCollision` | Skip the validation check at Tenant level for colliding Ingress hostnames | `false`
 `manager.image.repository` | Set the image repository of the controller. | `quay.io/clastix/capsule`
 `manager.image.tag` | Overrides the image tag whose default is the chart. `appVersion` | `null`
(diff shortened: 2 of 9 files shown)


MaxFedotov:
Hi Dario! Thanks a lot for the great work! 
I had a few additional issues, which I wanted to discuss regarding ingress hosts collision cases.

Let's also assume, that we are using some type of automatization, in order to publish ingress hosts to dns automatically (like https://github.com/kubernetes-sigs/external-dns/)

As a tenant-owner of `gas`, I create ingress with the host like `- host: "*.bigorg.com"`. That can lead to big problems for all other tenants, which will create ingresses like `oil.bigorg.com`.

But on the other hand, guys from `gas` tenant can buy a domain like `mygas.com`, which will be used only by them. And they want to create a single ingress in order to handle all requests to `*.mygas.com`. 

So there should be a list of domains for tenant, where wildcard ingress host may be created. All other wildcard ingress hosts should be denied by a webhook (if it is enabled)

bsctl (commented):
@prometherion this PR actually addresses #215 not #206. The latter and #207 are more related to the tenant validation not to the ingress validation.

bsctl:
change the default value to `true` according to this comment https://github.com/clastix/capsule/issues/215#issuecomment-791559641

prometherion:
Thanks @bsctl, just updated the main comment and updated the code with the right one! 😅

prometherion:
@MaxFedotov thanks for your insights! Just created a new issue for that so we can track it down properly.

bsctl (approved):
@prometherion 
thanks for your work, it lgtm

PR #216, opened 2021-03-05
Ingress hostname collision

### LINKED ISSUE #215: Prevent ingress hostname collision
# Describe the feature
Currently it is possible to create multiple ingresses with same ingress hostname across different tenants and within the same tenant as well.

A new CLI flag `--allow-ingress-hostname-collision` is added, defaulted to `true` to reflect current behaviour and do not break things. When set to `false`, Capsule will check if the Ingress hostname is already assigned to another ingress across different tenants and within the same tenant as well.

The check happens only in the Namespace resources managed by Capsule.

# Expected behavior
A configurable way to avoid ingress hostname collision

prometherion:
Closes #215.

By default, allowing Ingress hostname collision since it's the actual behavior of Capsule: for 0.1.0 we can change this making it breaking backed by the minor release.

### DIFF
--- a/charts/capsule/Chart.yaml
+++ b/charts/capsule/Chart.yaml
@@ -21,7 +21,7 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.0.12
+version: 0.0.13
 
 # This is the version number of the application being deployed.
 # This version number should be incremented each time you make changes to the application.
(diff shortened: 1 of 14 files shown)


bsctl (commented):
Helm Chart needs to be updated to include the new switch.

prometherion:
Going to merge this since CI today is totally flaky: tested locally on kind and it worked.

PR #217, opened 2021-03-05
Fix dirty flag on Capsule builds

### LINKED ISSUE #213: Capsule builds are always .dirty
We're having this output log that is useful when debugging Capsule to understand if the binary has been built with noncommitted changes that could lead to issues.

https://github.com/clastix/capsule/blob/2ea36db5d6a38cf15d777c2e34e4f9494ed0664b/main.go#L68

This variable is set at build-time, as following.

https://github.com/clastix/capsule/blob/2ea36db5d6a38cf15d777c2e34e4f9494ed0664b/Dockerfile#L21-L32

However, all the images are marked with `dirty` field, even the ones built using clean `git status` or through the CI system.

```json
{"level":"info","ts":"2021-03-04T07:31:44.732Z","logger":"setup","msg":"Capsule Version v0.0.5-rc1 737b6ce.dev.dirty"}
```

davideimola:
Fixes #213 
Hi, this will fix the dirty flag on the builds, I have moved out the `git diff` from the Dockerfile because we do not copy all the things inside the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,13 @@
 # Build the manager binary
 FROM golang:1.13 as builder
 
+ARG GIT_HEAD_COMMIT
+ARG GIT_TAG_COMMIT
+ARG GIT_LAST_TAG
+ARG GIT_MODIFIED
+ARG GIT_REPO
+ARG BUILD_DATE
+
 WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -19,16 +24,10 @@ COPY .git .git
 
 # Build
 RUN git config --get remote.origin.url > /tmp/GIT_REPO && \
```
prometherion, line 26: Can we go with `ARG` strategy also for `GIT_REPO` and `BUILD_DATE`?

davideimola: Yes, I think I can do it! :smile:

prometherion: It would be great, thanks!

Dockerfile
```diff
@@ -18,17 +25,9 @@ COPY pkg/ pkg/
 COPY .git .git
```
prometherion, line 25: I think we can drop the `.git` folder, here!

prometherion (changes requested):
Final request change, then we can merge although CI is red due to issues with GH.

Dockerfile
davideimola: @prometherion done it!

prometherion (approved):

Issue #229, opened 2021-03-09
Add a create user script for OCP 4.x (Experimental)


ValentinoUberti:
# Describe the feature

Create a script like hack/create-user.sh for OCP

# What would the new user story look like?

How would the new interaction with Capsule look like? E.g.

1. The cluster admin runs a create-user script for OCP

# Expected behavior
A demo user and tenant are created in OCP

bsctl:
closed by #230

PR #230, opened 2021-03-09
add: User script for Openshift (tested on OCP 4.7.1)

ValentinoUberti:
Issue #229
Added hack/create-user-openshift.sh for creating a demo user and tenant on OCP

### DIFF
--- a/hack/create-user-openshift.sh
+++ b/hack/create-user-openshift.sh
@@ -4,7 +4,7 @@
 # certificate signed by the Kubernetes CA itself.
 # It requires cluster admin permission.
 #
-# e.g.: ./create-user.sh alice oil
+# e.g.: ./create-user-openshift.sh alice oil
 #       where `oil` is the Tenant and `alice` the owner
 
 # Check if OpenSSL is installed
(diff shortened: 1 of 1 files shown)


hack/create-user-openshift.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+# This script uses Kubernetes CertificateSigningRequest (CSR) to a generate a
+# certificate signed by the Kubernetes CA itself.
+# It requires cluster admin permission.
+#
+# e.g.: ./create-user.sh alice oil
```
prometherion, line 7: Command name should be `create-user-openshift.sh`

prometherion (changes requested):
Just a small remark, once solved LGTM and thanks for testing Capsule on OCP! 🙏

hack/create-user-openshift.sh
ValentinoUberti: Fixed, thank you

prometherion (approved):
Going to merge once tests are green!

Issue #234, opened 2021-03-16
Add new group as an exception of the capsule-user-group


stg-0:
Hi everyone!

I've found an issue using Capsule in a (IMHO) non-conventional way, that's why I think this is a feature request and not a bug.

I've deployed Capsule on my k8s cluster, and used system:authenticated as the capsule-user-group so no user in Kubernetes remains outside the multi-tenancy, but when I tried to delete a Namespace with a NetworkPolicy created by the Tenant, I got an error code 500.

As far as I can tell, this error was caused by the AdmissionRequest for NetworkPolicy deletion sent by the controller (username: system:serviceaccount:kube-system:namespace-controller) which doesn't have a Name (referenced by the AdmissionRequest.Name), so the Get on the generic function cannot find the Resource.

I've fixed that error getting the Name from the OldObject in the AdmissionRequest, but then I got an error code 403.

Trying to understand this error, I've found out that Capsule's validating webhook allows requests for users outside the capsule-user-group without applying any filtering, but the request from the controller has the system:authenticated group I've configured.

So I came up with this idea for my use case: what if Capsule had an extra parameter to indicate a "system group" so a request with that group is considered "outside of Capsule" ?

That way, I could easily specify for example --system-user-group=system:serviceaccounts:kube-system and keep the --capsule-user-group=system:authenticated.

What do you think?

prometherion:
First of all, thanks for giving Capsule a try! 🤗 

> I've deployed Capsule on my k8s cluster, and used system:authenticated as the capsule-user-group so no user in Kubernetes remains outside the multi-tenancy

I'd say this, although non documented, it's a bad practice: the issue is that RBAC is used by several components in Kubernetes, such as `system:node` used by the Kubelet certificate, etc. Multi-tenancy should be applied just to a small set of users and the `system:authenticated` group is a too much broader audience, bringing side-effects or issuess as the one as you just faced.

> what if Capsule had an extra parameter to indicate a "system group" so a request with that group is considered "outside of Capsule" ?

Using the additional CLI parameter to exclude users from the validation would be bad since it smells like a workaround for a code that is malformed.

Besides this clarification, I think we could refactor a little bit the middleware we're using here:

https://github.com/clastix/capsule/blob/56adfe6a35d2271f1556d794614007aee656dbb5/main.go#L176-L178

Instead of relying on the group belonging, we could perform these tasks:

- checking if the resource (NetworkPolicy for the said case) is managed by Capsule
- if it's not, passthrough
[...]

PR #235, opened 2021-03-16
Add system-user-group as exception of capsule-user-group

stg-0:
Hi there!

I've created this issue explaining this PR with words and the use-case I'm trying to solve: #234

In short, I would like to include a parameter such as system-user-group to avoid filtering requests from users belonging to that group, even if the user belongs to the capsule-user-group.


stg-0:
Superseded by #237. Thanks @prometherion!

PR #236, opened 2021-03-16
Include capsule-proxy in main Capsule Chart

### LINKED ISSUE #231: Install the capsule-proxy service along with capsule-controller-manager by main Capsule Helm Chart
# Describe the feature
It would be nice to install the `capsule-proxy` service along with `capsule-controller-manager` by the main Capsule Helm Chart. ~We use the sub charts feature of Helm.~

As cluster admin, install the capsule-proxy along with capsule-controller-manager by the same Helm Chart

# Expected behavior
see above

bsctl:
close #231

### DIFF
--- a/hack/create-user-openshift.sh
+++ b/hack/create-user-openshift.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+# This script uses Kubernetes CertificateSigningRequest (CSR) to a generate a
+# certificate signed by the Kubernetes CA itself.
+# It requires cluster admin permission.
+#
+# e.g.: ./create-user-openshift.sh alice oil
+#       where `oil` is the Tenant and `alice` the owner
+
+# Check if OpenSSL is installed
+if [[ ! -x "$(command -v openssl)" ]]; then
+    echo "Error: openssl not found"
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


bsctl:
@prometherion @MaxFedotov please review. This PR adds the capsule-proxy helm chart into the main chart. This actually makes the capsule-proxy helm chart redundant since the capsule-proxy component itself cannot be installed as standalone service without the capsule-manager. Should we deprecate the capsule-proxy helm chart?

bsctl:
We do not need to add dependancy charts since they can be pulled with the command:

`$ helm dependency build capsule`

It pulls the dependancy charts listed in the `Chart.yaml` file:

```
...
dependencies:
- name: capsule-proxy
  version: "0.1.1"
  repository: https://clastix.github.io/charts
  condition: capsule-proxy.enabled
```

so we can have the `capsule-proxy` chart in one single place.

bsctl:
e2e tests are broken since we need to run `helm dependency build capsule` before to deploy

```
Error: found in Chart.yaml, but missing in charts/ directory: capsule-proxy
helm.go:94: [debug] found in Chart.yaml, but missing in charts/ directory: capsule-proxy
```

Issue #241, opened 2021-03-24
Namespaces within a Tenant cannot be selected in NetworkPolicies


stg-0:
Hi everyone!

I'm trying the brand new v0.0.5 release and found out that, due to the issue #200, there's no way to reference (using namespaceSelector) a particular Namespace within a tenant in a NetworkPolicy.

As stated in the mentioned issue and in the [official documentation](https://kubernetes.io/docs/concepts/services-networking/network-policies/#what-you-can-t-do-with-network-policies-at-least-not-yet), targeting of namespaces by name in NetworkPolicies is not possible and, with the added functionality, there's no way to tag each Namespace separately.

To be able to reference each Namespace in a NetworkPolicy, a possible solution would be to automatically tag each Namespace with a label with its own name.

Something like this:

```
apiVersion: v1
kind: Namespace
metadata:
  labels:
    capsule.clastix.io/tenant: mytenant
    name: mytenant-ns
  name: mytenant-ns
```

prometherion:
Looks good to me: @bsctl @MaxFedotov any remark on this?

I also think this can be a really good first issue, any contribution is highly appreciated! ❤️

MaxFedotov:
We are already doing this in order to support Calico `GlobalNetworkPolicies`.
But we are using `namespacesMetadata.additionalLabels` in Tenant spec:
```yaml
  namespacesMetadata:
    additionalLabels:
      capsule.clastix.io/tenant: {tenant_name}
```

stg-0:
I need to identify each Namespace within a tenant separately, not all the Namespaces belonging to a particular tenant.

prometherion:
I think the use-case here is different: we don't have to replicate the tenant name but, rather, the Namespace name. Labeling a Namespace is either disabled due to the reported bug/enhancement, we have just to rely on the `namespacesMetadata.additionalLabels` but these are going to be applied across all the Tenant Namespace resources.

I guess the need for this feature is to take advantage of the Capsule NetworkPolicy resources reconciliation: these ones cannot either be deleted by the Tenant Owner and ensure to select the single Namespace and isolate itself rather than at Tenant level.

Please, @stg-0, can you confirm this or, rather, provide a sort of user-story for this enhancement?

stg-0:
[...]

PR #242, opened 2021-03-24
Add name label to each Namespace

stg-0:
Hi there!

I've created this issue explaining this PR with words and the use-case I'm trying to solve: #241 

In short, I would like to include a label called "name" (or any other suggested key) per Namespace, so I can reference them in NetworkPolicies.


prometherion (approved):

prometherion:
@MaxFedotov I think we can easily merge this, it's just a matter of adding a label: WDYT?

MaxFedotov:
> @MaxFedotov I think we can easily merge this, it's just a matter of adding a label: WDYT?

Agreed, lgtm!

PR #259, opened 2021-05-19
feat: block use of NodePort Services

### LINKED ISSUE #249: Block use of NodePort Services
# Describe the feature
It would be nice to leave the cluster admin to block the use of `NodePort` services for a given tenant. A `NodePort` services created in a given tenant will open a port on all the worker nodes of the cluster. In some multi-tenant scenarios, for example tenants with dedicate node pools, this could be an issue since the port is opened on all the nodes, including nodes belonging to other tenants.

This feature is also required to get compliance with the Multi-tenancy Kubernetes Working Group benchmark, see #68 

1. The cluster admin creates a tenant with `.spec.enableNodePorts=false`
4. Tenant owner creates a kubernetes service with `.spec.type=NodePort` 
5. A validation webhooks intercepts the requests and denies it. A proper error message is displayed.

# Expected behavior
The cluster admin can prevent some tenants to create services with `.spec.type=NodePort`

MaxFedotov:
closes #249

### DIFF
--- a/pkg/webhook/services/validating.go
+++ b/pkg/webhook/services/validating.go
@@ -32,6 +32,10 @@ import (
 
 // +kubebuilder:webhook:path=/validating-external-service-ips,mutating=false,sideEffects=None,admissionReviewVersions=v1,failurePolicy=fail,groups="",resources=services,verbs=create;update,versions=v1,name=validating-external-service-ips.capsule.clastix.io
 
+const (
+	enableNodePortsAnnotation = "capsule.clastix.io/enable-node-ports"
+)
+
 type webhook struct {
 	handler capsulewebhook.Handler
 }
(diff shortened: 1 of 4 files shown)


pkg/webhook/services/validating.go
```diff
@@ -93,6 +97,10 @@ func (r *handler) handleService(ctx context.Context, clt client.Client, decoder
 		}
 	}
 
+	if svc.Spec.Type == corev1.ServiceTypeNodePort && tnt.GetAnnotations()[enableNodePortsAnnotation] == "false" {
```
prometherion, line 100: Don't we have to check if `tnt.GetAnnotations()` is not nil here to avoid nil pointer dereference?

MaxFedotov: i've got a test case for this - https://github.com/clastix/capsule/pull/259/files#diff-4e5ed4f1f8da5436910e7528461ba68f9ceb5eb373755b3cdb857856c7e36d1dR33-R44
and everything seems ok :)

PR #260, opened 2021-05-19
feat: use multiple groups as capsule-user-group

### LINKED ISSUE #258: Use multiple groups as capsule-user-group
# Describe the feature
Currently, only user requests belonging to the group specified in `--capsule-user-group` are managed by Capsule. In some cases, there are users in different groups that need to use Capsule and it is not always possible to create a new dedicated common group and move all the existing users into it (think a large legacy AD/LDAP user base managed by a corporate security team).

It would be nice to specify a list of groups as `--capsule-user-groups=capsule.clastix.io,engineering,sudoers,sysadmins`.

How would the new interaction with Capsule look like? E.g.

1. What are the prerequisites for this?
4. All the magic happens in the background

Feel free to add a diagram if that helps explain things.

# Expected behavior
User requests belonging to users in the specified groups `--capsule-user-groups=` are managed by the Capsule Controller.

MaxFedotov:
closes #258

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
             ${{ runner.os }}-
       - run: make manifests
       - name: Checking if manifests are disaligned
-        run: test -z "$(git diff 2> /dev/null)"
+        run: if [[ $(git diff | wc -l) -gt 0 ]]; then echo ">>> Untracked generated files have not been committed" && git --no-pager diff && exit 1; fi
       - name: Checking if manifests generated untracked files
         run: test -z "$(git ls-files --others --exclude-standard 2> /dev/null)"
       - run: make fmt vet
--- a/controllers/rbac/manager.go
+++ b/controllers/rbac/manager.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 
+	b64 "encoding/base64"
+
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/go-multierror"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -34,10 +36,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	rbacLabel = "capsule.clastix.io/rbac"
+)
+
 type Manager struct {
-	CapsuleGroup string
-	Log          logr.Logger
-	Client       client.Client
+	CapsuleGroups []string
+	Log           logr.Logger
+	Client        client.Client
 }
 
 // Using the Client interface, required by the Runnable interface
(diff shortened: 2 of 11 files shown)


prometherion:
It seems we got some untracked files upon code generation: can you check it, please?

pkg/webhook/utils/in_capsule_group.go
```diff
@@ -47,7 +47,12 @@ func (h handler) isCapsuleUser(req admission.Request) bool {
 	if groupList.Find("system:serviceaccounts:kube-system") {
 		return false
 	}
-	return groupList.Find(h.capsuleGroup)
+	for _, group := range h.capsuleGroup {
```
prometherion, line 50: It's not optimal here, were ending with `m x O(log n)` here, but at the same time I think it's a pay-off: if you can map users in a single group it works at its best, otherwise prepare for sole bad performances.

pkg/webhook/utils/in_capsule_group.go
```diff
@@ -26,15 +26,15 @@ import (
...
+func InCapsuleGroup(capsuleGroup []string, webhookHandler webhook.Handler) webhook.Handler {
 	return &handler{
 		handler:      webhookHandler,
 		capsuleGroup: capsuleGroup,
 	}
 }
 
 type handler struct {
-	capsuleGroup string
+	capsuleGroup []string
```
prometherion, line 37: Just a naming issue: maybe better `capsuleGroups`?

pkg/webhook/utils/in_capsule_group.go
```diff
@@ -26,15 +26,15 @@ import (
 	"github.com/clastix/capsule/pkg/webhook"
 )
 
-func InCapsuleGroup(capsuleGroup string, webhookHandler webhook.Handler) webhook.Handler {
+func InCapsuleGroup(capsuleGroup []string, webhookHandler webhook.Handler) webhook.Handler {
 	return &handler{
```
prometherion, line 30: `func InCapsuleGroups`?

controllers/rbac/manager.go
```diff
@@ -35,7 +37,7 @@ import (
 )
 
 type Manager struct {
-	CapsuleGroup string
+	CapsuleGroup []string
```
prometherion, line 40: `CapsuleGroups`?

prometherion (commented):
I saw you updated also the Cluster Role binding allowing the Namespace creations to groups, but I guess we have to update the Helm chart postDelete hook, as well the `make destroy` target, isn't it?

I think we can easily address it using deletion of resources based on labels, so we just need to put a sort of unique labels so we can target them and delete easily: WDYT?

pkg/webhook/utils/in_capsule_group.go
MaxFedotov: Yep, thank you. Renamed

pkg/webhook/utils/in_capsule_group.go
MaxFedotov: Done

controllers/rbac/manager.go
MaxFedotov: Done

MaxFedotov:
> It seems we got some untracked files upon code generation: can you check it, please?

Can't understand what the problem is. When I ran `make manifests` locally - I don't have any new or untracked files :(

prometherion:
> Can't understand where the problem is. When I ran `make manifests` locally - I don't have any new or untracked files :(

[We updated the `controller-gen` binary to v0.5.0 in order to support `admission.k8s.io/v1` and `apiextensions.k8s.io/v1`](https://github.com/clastix/capsule/pull/255). Not sure this is related, BTW.

prometherion:
Anyway, let's try to debug it from GH Actions: I would change the `test -z "$(git diff 2> /dev/null)"` step to something like this:

```bash
if [[ $(git diff | wc -l) -gt 0 ]]; then
  echo ">>> Untracked generated files have not been committed"
  git diff
  exit 1
fi
```

MaxFedotov:
@prometherion 
```
>>> Untracked generated files have not been committed
diff --git a/go.mod b/go.mod
index b8b85ab..8c90ae0 100644
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.10.2
 	github.com/pkg/errors v0.9.1
-	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
 	go.uber.org/zap v1.15.0
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
```
lol, it was due to pflag // indirect reference, which was added by my goland IDE

MaxFedotov:
@prometherion can't understand what's going on with tests, they are very flappy

prometherion:
> @prometherion can't understand what's going on with tests, they are very flappy

It happens from time to time, unfortunately.

Regarding the single commit, may I ask you to split the changes into smaller chunks and according to the topic? For example, the changes to the Helm chart could be `build(helm): <commit_message>`.

charts/capsule/Chart.yaml
```diff
@@ -21,7 +21,7 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.0.19
+version: 0.0.20
```
prometherion, line 24: Let's avoid incrementing the Chart version since the proposed changes must be addressed when releasing a new Capsule release and would break up the Helm release during the clean-up.

.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
             ${{ runner.os }}-
       - run: make manifests
       - name: Checking if manifests are disaligned
-        run: test -z "$(git diff 2> /dev/null)"
+        run: if [[ $(git diff | wc -l) -gt 0 ]]; then echo ">>> Untracked generated files have not been committed" && git --no-pager diff && exit 1; fi
```
prometherion, line 40: Also, this change should be address in a separate commit, something as `chore(ci): output diff files for manifests files`

controllers/rbac/manager.go
```diff
@@ -115,36 +121,42 @@ func (r *Manager) Reconcile(ctx context.Context, request reconcile.Request) (res
...
-			{
-				Kind: "Group",
-				Name: r.CapsuleGroup,
+	for _, group := range r.CapsuleGroups {
+		name := b64.RawStdEncoding.EncodeToString([]byte(group))
+		crb := &rbacv1.ClusterRoleBinding{
+			ObjectMeta: v1.ObjectMeta{
+				Name: fmt.Sprintf("%s-%v", ProvisionerRoleName, name),
+				Labels: map[string]string{
+					rbacLabel: "",
```
prometherion, line 130: What about to use `capsule.clastix.io/rbac` as key and the Object name as value?

.github/workflows/ci.yml
MaxFedotov: done

controllers/rbac/manager.go
MaxFedotov: done

charts/capsule/Chart.yaml
MaxFedotov: ok, reverted

.github/workflows/ci.yml
MaxFedotov: done

prometherion (approved):

PR #265, opened 2021-05-27
fix: wrong order of checks in validating-external-service-ips webhook

### LINKED ISSUE #249: Block use of NodePort Services

MaxFedotov:
closes #249

### DIFF
--- a/pkg/webhook/services/validating.go
+++ b/pkg/webhook/services/validating.go
@@ -68,10 +68,6 @@ func (r *handler) handleService(ctx context.Context, clt client.Client, decoder 
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	if svc.Spec.ExternalIPs == nil {
-		return admission.Allowed("")
-	}
-
 	tntList := &v1alpha1.TenantList{}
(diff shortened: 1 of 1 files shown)


prometherion (approved):
Webhooks are getting a mess, we have to start thinking refactoring them.

But this is not the time, it's a LGTM (for now).

PR #269, opened 2021-05-30
CapsuleConfiguration CRD

### LINKED ISSUE #122: Capsule configuration via a CRDs
With Capsule growing and getting more features, probably we should start thinking about providing an effective way for iterating with Capsule options rather than through CLI flags.

Ingress Controller moved toward a CRDs thanks to Ingress Class (especially, the `.spec.parameters` key as a typed local object reference).

Although this would be nice and cool, wondering if we really need a CRDs: Capsule is expected to run in an exclusive way on the Kubernetes cluster, so that would mean having just a single resource of the custom type and sounds odd.

prometherion:
Closes #122.

Rather than dealing with CLI flags, a new CRD is introduced.

```yaml
kind: CapsuleConfiguration
metadata:
  name: default
spec:
  userGroups: ["capsule.clastix.io"]
  forceTenantPrefix: false
  protectedNamespaceRegex: ""
  allowTenantIngressHostnamesCollision: false
  allowIngressHostnameCollision: false
```

All the options from CLI have been hardly deprecated, substituted with the new CRD reference (`--configuration-name=default`).

This is absolutely a breaking change but it's required to ship the 0.1.0 release.

We should have also good news regarding e2e testing since the configuration reloads is at runtime, allowing to skip the Pod respawning.


prometherion:
Tests are running in ~6.30 minutes, we halved the average execution! 😎 🚀

bsctl:
@prometherion amazing work.

MaxFedotov:
@prometherion wow, that's great job!

config/samples/capsule_v1alpha1_capsuleconfiguration.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
bsctl, line 1: installing with `make deploy` but it seems the default configuration is not created

```
$ kubectl get capsuleconfigurations
No resources found in capsule-system namespace.
$ kubectl get capsuleconfigurations -A
No resources found
```

but it is referenced in the manager

```yaml
...
    Command:
      /manager
    Args:
      --enable-leader-election
      --zap-encoder=console
      --zap-log-level=debug
      --configuration-name=default
```

config/samples/capsule_v1alpha1_capsuleconfiguration.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: capsule.clastix.io/v1alpha1
+kind: CapsuleConfiguration
+metadata:
+  name: default
+spec:
+  userGroups: ["capsule.clastix.io"]
+  forceTenantPrefix: false
+  protectedNamespaceRegex: ""
+  allowTenantIngressHostnamesCollision: false
```
bsctl, line 10: it should be `allowTenantIngressHostnamesCollision: false` according to #207

docs/operator/references.md
```diff
@@ -671,13 +671,37 @@ Option | Description | Default
...
+
+Option | Description | Default
+--- | --- | ---
+`.spec.forceTenantPrefix` | Force the tenant name as prefix for namespaces: `<tenant_name>-<namespace>`.  | `false`
+`.spec.userGroups` | Array of Capsule groups to which all tenant owners must belong. | `[capsule.clastix.io]`
+`.spec.protectedNamespaceRegex` | Disallows creation of namespaces matching the passed regexp. | `null`
+`.spec.allowTenantIngressHostnamesCollision` | By default, Capsule allows Ingress hostname collision: set to `false` to enforce this policy. | `true`
+`.spec.allowIngressHostnameCollision` | Toggling this, Capsule will not check if a hostname collision is in place, allowing the creation of two or more Tenant resources although sharing the same allowed hostname(s). | `false`
+
+Upon installation using Kustomize or Helm, a `default` resource will be created.
```
bsctl, line 703: by changing the configuration at runtime, do we expect Capsule to automatically reloads the new configuration?

bsctl: > by changing the configuration at runtime, do we expect Capsule to automatically reloads the new configuration?

yep, it happens! Hurray!
However, it would be nice to have track of this event in the logs since it's a main change in the controller behaviour.

config/samples/capsule_v1alpha1_capsuleconfiguration.yaml
prometherion: Solved with commit 537e4a3

config/samples/capsule_v1alpha1_capsuleconfiguration.yaml
prometherion: I'd like to address this in a separate PR, referencing this issue since it would require some test changes and wanted to avoid making a fat PR.

docs/operator/references.md
prometherion: Addressed with 4945411

bsctl (approved):
@prometherion LGTM!!!

PR #281, opened 2021-06-04
Docker-ci for armv7 / arm64 support 

### LINKED ISSUE #244: Support for Arm-based Docker images
# Describe the feature

Actually, all the Docker images are built using the `amd64` architecture.

https://github.com/clastix/capsule/blob/12237ae106d6e9bbcb49e1091318534a901f29cd/Dockerfile#L27

Capsule should be runnable also on Arm and Arm64-based infrastructures since there are plenty of projects running Kubernetes at the edge.

1. Definitely, we have to switch from quay.io based Docker image creation to GitHub Actions, requiring to provide a new GitHub Secret to allow push of built Docker images
2. Using [`docker buildx`](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) to allow the creation of multi-arch images
3. Evaluate if we need a hub.docker.com account

# Expected behavior

Running `clastix/capsule:v0.0.5` (YMMV) container image on an Arm-based Kubernetes cluster, Capsule should be up and running.

ptx96:
Closes #244

- Added armv7 / arm64 support (k8s deprecated armv6)
- Automated docker buildx & push to clastix quay.io registry using github actions
- Added proprietary kubectl image from quay.io

Completely tested on amd64 / arm64

### DIFF
--- a/.github/workflows/docker-ci.yml
+++ b/.github/workflows/docker-ci.yml
@@ -1,0 +1,73 @@
+name: docker-ci
+
+on:
+  push:
+    tags:        
+      - "v*"
+
+jobs:
+  docker-ci:
+    runs-on: ubuntu-20.04
+    steps:
+
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+             quay.io/${{ github.repository }}
+          tags: |
+            type=semver,pattern={{raw}}
+          flavor: |
+            latest=false
+
+      -
+        name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: arm64,arm
+
+      -
+        name: Set up Docker Buildx
+        id: buildx
+        with:
+          install: true
+        uses: docker/setup-buildx-action@v1
+
+      -
+        name: Inspect builder
+        run: |
+          echo "Name:      ${{ steps.buildx.outputs.name }}"
+          echo "Endpoint:  ${{ steps.buildx.outputs.endpoint }}"
+          echo "Status:    ${{ steps.buildx.outputs.status }}"
+          echo "Flags:     ${{ steps.buildx.outputs.flags }}"
+          echo "Platforms: ${{ steps.buildx.outputs.platforms }}"
+
+      -
+        name: Login to quay.io Container Registry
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Build the manager binary
-FROM quay.io/bitnami/golang:1.16 as builder
+FROM golang:1.16 as builder
```
prometherion, line 2: This switch back to Docker Hub repository is due to missing architectures in the quay.io hosted one.

.github/workflows/docker-ci.yml
```diff
@@ -0,0 +1,73 @@
+name: docker-ci
+
+on:
+  push:
+    tags:        
+      - '*'
```
prometherion, line 6: This would collide with the Helm action tag: https://github.com/clastix/capsule/blob/7c1592e739838db822b317b0a6a356e39e2769c6/.github/workflows/helm.yml#L20

Can we just trigger this workflow when the tag starts with `v`?

Not sure if this would work:

```yaml
on:
  push:
    tags:
      - "v*"
```

prometherion (changes requested):
Thanks for the big effort on this, really thrilled for landing this to Capsule!

Just a single remark regarding the watched tags for the `docker-ci` workflow, but overall LGTM!

prometherion:
Regarding the `quay.io/clastix.kubectl` Image, may I ask you to share the Dockerfile and all the related files to the repository https://github.com/clastix/kubectl?

.github/workflows/docker-ci.yml
ptx96: this line has been changed as requested inside commit [8473750](https://github.com/clastix/capsule/commit/8473750da9f362a4a74b76a5152d4f5d72648127)

everything works well

ptx96:
> Just a single remark regarding the watched tags for the `docker-ci` workflow, but overall LGTM!

I've fixed it with last commit; docker-ci trigger works like a charm!

> Regarding the `quay.io/clastix.kubectl` Image, may I ask you to share the Dockerfile and all the related files to the repository https://github.com/clastix/kubectl?

Sure; I'll also open an issue to automate clastix/kubectl release with gh actions (as done here).

> Thanks for the big effort on this, really thrilled for landing this to Capsule!

Thank you @prometherion, I'm eager to see v0.1.0 tag!

prometherion (approved):

PR #287, opened 2021-06-12
Image PullPolicy enforcement for v1alpha

### LINKED ISSUE #271: Require tenants to use imagePullPolicy as Always
# Describe the feature
The Cluster admin can require tenant owners to always use `imagePullPolicy: Always` so that their private images can only be used by those who have the credentials to pull them.

This is also required for SIG multi-tenancy [benchmarks](https://github.com/kubernetes-sigs/multi-tenancy/tree/master/benchmarks) as part of L1 profile.

1. the cluster admin creates a tenant manifest as:
```yaml
kind: Tenant
metadata:
  name: oil
spec:
  owner:
    name: alice
    kind: User
  allowedImagePullPolicy:
  - Always
  - Never
  - IfNotPresent
```
4. Tenant owner can only creates pods with allowed `ImagePullPolicy`. When not specified, any value is admitted, including `null`, i.e. defaulted to `Always` if `:latest` tag is specified, or `IfNotPresent` otherwise.

# Expected behavior
see above

prometherion:
Closes #271.

### DIFF
--- a/api/v1alpha1/domain/imagepullpolicy.go
+++ b/api/v1alpha1/domain/imagepullpolicy.go
@@ -1,0 +1,51 @@
+// Copyright 2020-2021 Clastix Labs
+// SPDX-License-Identifier: Apache-2.0
+
+package domain
+
+import (
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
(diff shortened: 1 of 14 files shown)


bsctl (approved):
LGTM @prometherion thanks

PR #300, opened 2021-06-24
Created Grafana Dashboard and related charts/manifests

### LINKED ISSUE #150: Provide a Grafana dashboard for metrics overview
# Describe the feature

Having a Grafana dashboard JSON would be great in order to fetch some stats: metrics are already provided by the Controller Runtime package and available at port 8080 on `/metrics` endpoint, just need to puzzle them.

Along with this, a `ServiceMonitor` instance must be declared in the Helm Chart and Kustomize manifests.

**As a**  Cluster Administror
**I want to** have a Grafana Dashboard
**so that I can** get a clear picture of the latency on available controllers, total amount of requeued items due to error, etc.

# Expected behavior

Once the JSON is there, just upload to Grafana and jazz hands for happy monitoring of Capsule.

ptx96:
Closes #150

### DIFF
--- a/config/prometheus/kustomization.yaml
+++ b/config/prometheus/kustomization.yaml
@@ -1,2 +1,4 @@
 resources:
 - monitor.yaml
+- role.yaml
+- rolebinding.yaml
--- a/config/prometheus/monitor.yaml
+++ b/config/prometheus/monitor.yaml
@@ -1,16 +1,18 @@
-
 # Prometheus Monitor Service (Metrics)
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
     control-plane: controller-manager
-  name: controller-manager-metrics-monitor
+  name: capsule-monitor
   namespace: system
 spec:
   endpoints:
-    - path: /metrics
+    - interval: 15s
+      path: /metrics
       port: metrics
+  jobLabel: controller-manager
+  namespaceSelector:
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/config/prometheus/role.yaml
+++ b/config/prometheus/role.yaml
@@ -1,0 +1,18 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: capsule-metrics-role
+  namespace: capsule-system
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
--- a/config/prometheus/rolebinding.yaml
+++ b/config/prometheus/rolebinding.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: capsule-metrics-rolebinding
+  namespace: system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: capsule-metrics-role
+subjects:
+- kind: ServiceAccount
+  name: capsule
+  namespace: capsule-system
--- a/charts/capsule/README.md
+++ b/charts/capsule/README.md
@@ -88,6 +88,9 @@ Parameter | Description | Default
 `replicaCount` | Set the replica count for Capsule pod. | `1`
 `affinity` | Set affinity rules for the Capsule pod. | `{}`
 `podSecurityPolicy.enabled` | Specify if a Pod Security Policy must be created. | `false`
+`serviceMonitor.enabled` | Specify if a Service Monitor must be created. | `false`
+`serviceMonitor.serviceAccount.name` | Specify Service Account name for metrics scrape. | `capsule`
+`serviceMonitor.serviceAccount.namespace` | Specify Service Account namespace for metrics scrape. | `capsule-system`
 
 ## Created resources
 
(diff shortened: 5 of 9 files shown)


config/prometheus/kustomization.yaml
```diff
@@ -1,2 +1,4 @@
 resources:
 - monitor.yaml
+- role.yaml
+- rolebinding.yaml
```
prometherion, line 4: Missing new line at the end of the file.

config/prometheus/monitor.yaml
```diff
@@ -1,16 +1,20 @@
...
 # Prometheus Monitor Service (Metrics)
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
     control-plane: controller-manager
-  name: controller-manager-metrics-monitor
-  namespace: system
+  name: capsule-monitor
+  namespace: capsule-system
```
prometherion, line 8: `Namespace` must remain to `system` since `kustomize` will change it at installation time.

config/prometheus/monitor.yaml
```diff
@@ -1,16 +1,20 @@
...
 spec:
   endpoints:
-    - path: /metrics
+    - interval: 15s
+      path: /metrics
       port: metrics
+  jobLabel: app.kubernetes.io/name
+  namespaceSelector:
```
prometherion, line 15: Do we need the `namespaceSelector` here although the Service Monitor is living inside the Capsule namespace?

The hard thing is that we're not using Helm and I don't know how to inject the `namespace` value provided by the `config/default/kustomization.yaml` file.

config/prometheus/monitor.yaml
```diff
@@ -1,16 +1,20 @@
...
+    matchNames:
+    - capsule-system
   selector:
     matchLabels:
-      control-plane: controller-manager
+      app.kubernetes.io/name: capsule
```
prometherion, line 20: We have to use the labels provided by kustomize, not the Helm ones, so `control-plane=controller-manager` should be good.

config/prometheus/role.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - ""
+  resources:
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
```
prometherion, line 19: Why the Prometheus role must access the Ingress objects?

prometherion (changes requested):
Some changes are requested for the `kustomize` based installation.

Some remarks on the RBAC and the Pod selector.

prometherion:
@ptx96 may I ask you to share also a screenshot of the dashboard?

Would be great having it in the documentation.

config/prometheus/monitor.yaml
ptx96: `namespaceSelector` should not be related to service monitor location ( maybe you're thinking about `ServiceMonitorNamespaceSelector` );
nevertheless, can be left empty without any problems.

Keep in mind that, in this way, servicemonitor will search for services to scrape (by matchLabels) in every namespace.

config/prometheus/monitor.yaml
ptx96: checked "capsule/config/manager/metrics_service.yaml" and found :

`  labels:
    control-plane: controller-manager`

so, as you said, it should be good to revert this line

config/prometheus/monitor.yaml
prometherion: You're totally right.

I checked out what's the scaffolded ServiceMonitor by `operator-sdk create api` command and it's the following:

```yaml
# Prometheus Monitor Service (Metrics)
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    control-plane: controller-manager
  name: controller-manager-metrics-monitor
  namespace: system
spec:
  endpoints:
    - path: /metrics
      port: https
      scheme: https
      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
      tlsConfig:
        insecureSkipVerify: true
  selector:
    matchLabels:
      control-plane: controller-manager
```

Let's use this non-optimal solution for the ServiceMonitor backed by `kustomize`, we got more control over Helm.

config/prometheus/monitor.yaml
```diff
@@ -1,16 +1,18 @@
...
-  name: controller-manager-metrics-monitor
+  name: capsule-monitor
   namespace: system
 spec:
   endpoints:
-    - path: /metrics
+    - interval: 15s
+      path: /metrics
       port: metrics
+  jobLabel: app.kubernetes.io/name
```
prometherion, line 14: According to the Prometheus Operator APIs:

```
job label: The label to use to retrieve the job name from
```

Since this is the installation with `kustomize`, we have to use the value `controller-manager`.

prometherion (changes requested):

config/prometheus/rolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: capsule-metrics-rolebinding
+  namespace: capsule-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: capsule-metrics-role
+subjects:
+- kind: ServiceAccount
+  name: capsule
+  namespace: capsule-system
```
prometherion, line 15: must be set to `system` in order to let `kustomize` change it to the desired namespace.

prometherion (changes requested):

prometherion (approved):

PR #303, opened 2021-06-26
Using multiple handlers per single resource.

### LINKED ISSUE #297: Refactoring of webhooks
Since new features are added, we're noticing an increase in new webhooks.

These are hard to track and require a lot of gluing, especially for the `kustomize` integration since we have to add the `namespaceSelector` to decrease the pressure on the Capsule webserver.

We could use a single webhook per resource (`/services`, `/ingresses`, `/namespaces`, `/persistentvolumeclaims`, etc.) using a middleware chaining.

The benefit of this is that we don't need to track down multiple times new webhooks, both on Helm and Kustomize.

prometherion:
Closes #297.

Managing multiple webhooks was a pain and it's more convenient using a route and using a sort of middleware chain.

This is absolutely a breaking change, especially for the `kustomize` and `helm` installation modes.


bsctl (approved):
@prometherion LVGTM
Just made few manual tests on main use cases, assuming the automatic test suite already sanitised the changes. Thanks for the great work!

prometherion:
> assuming the automatic test suite already sanitised the changes. Thanks for the great work!

Absolutely, I have to admit the first changes were dramatic since I messed up some stuff but after the reorg, the e2e went smooth!

PR #310, opened 2021-06-28
WIP: New Tenant v1alpha2 version 

### LINKED ISSUE #286: New Tenant v1beta1 version
New features have been implemented in the current version, most of them are available using annotations that are just a workaround (or rather, a Technology Preview) for the said features.

We have to start releasing the new version that will provide real specification keys for the upcoming and planned features.

- [Assign Priority Classes to tenants](https://github.com/clastix/capsule/issues/257)
- [Require tenants to use imagePullPolicy as Always](https://github.com/clastix/capsule/issues/271)
- [Add applications to act as tenant owner](https://github.com/clastix/capsule/issues/276)
- [Make configurable the Resource Quota enforcement at tenant level](https://github.com/clastix/capsule/issues/50)

Luckily, we had a discussion on #202 and from the reconciliation PoV it's pretty easy, not the same for webhooks but we can work on that.

@MaxFedotov do you already have some rough ideas on the new API version structure?

MaxFedotov:
@prometherion this is the first commit for #286. Adds new v1alpha2 Tenant spec


prometherion (commented):
I'm seeing some duplicated structs used also on `v1alpha1`, such as `AllowedListSpec` and `AdditionalMetadata`, just to name a few of them.

WDTY abstracting these structures in the `domain` package so we can reuse them across the two types? tl;dr; DRY principle

MaxFedotov:
I was thinking about it, but then I checked https://github.com/Azure/azure-service-operator as an example, and they are copying structs between versions, not using abstract ones. And I think this approach had an advantage, because we will be able to modify them in future without breaking v1alpha1 CRDs.

prometherion (approved):
LGTM

PR #327, opened 2021-07-05
bug: ensuring to update the conversion webhook CA bundle

### LINKED ISSUE #326: Conversion webhook CA bundle is not updated when installing with Kustomize
Related to #321.

When installing with `kustomize`, the webhook CA bundle is not properly updated by the CA controller.

prometherion:
Closes #326.


MaxFedotov (approved):
lgtm!

PR #348, opened 2021-07-17
Single YAML file installer

### LINKED ISSUE #347: Providing a single YAML file installer
Actually, we support two kinds of installation:

- Helm
- Kustomize

Kustomize is still required for the development purpose since the Kubebuilder annotations don't provide all the required extra fields, but forcing end-users to have it as installation dependency is cumbersome.

Would be great having a single YAML file that is going to be used to install all the required resources, something as `kubectl apply -f https://raw.githubusercontent.com/clastix/capsule/master/config/install.yaml`.

With this, we're gaining also the advantage of tracking in a more efficient way the final manifests we're using to update the Helm charts, as well tracking uncommitted changes or diff during the CI.

prometherion:
Closes #347.

Pretty sure you're happy for this, @bsctl.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Cache Go modules
         uses: actions/cache@v1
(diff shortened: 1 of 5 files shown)


bsctl (approved):
lgtm @prometherion like single yaml installation

PR #350, opened 2021-07-19
Support for v1beta1 in capsule-proxy

### LINKED ISSUE #349: Support v1beta1 in capsule-proxy
In order to support v1beta1 API in capsule-proxy we need to add a few changes to capsule v1beta1 API:
1. Use constants in `owner` spec instead of strings (so when we import v1beta1 capsule API in capsule-proxy we can use the same constants)
2. Add additional function to get `proxySettings` for a specific tenant `owner`
3. Update `create-user.sh` script to support multiple `groups` as tenant `owner`

MaxFedotov:
closes #349


api/v1beta1/tenant_func.go
```diff
@@ -36,3 +36,12 @@ func (t *Tenant) AssignNamespaces(namespaces []corev1.Namespace) {
 	t.Status.Namespaces = l
 	t.Status.Size = uint(len(l))
 }
+
+func (t *Tenant) GetOwnerProxySettings(name string, kind OwnerKind) []ProxySettings {
```
prometherion, line 40: I don't want to be pedantic here but could you we use here a sorted slice in order to take advantage of the binary search?

MaxFedotov: done :)

prometherion (changes requested):
LGTM but I would ask you to split changes in two commits:

`chore` for the `hack/create-user.sh` (would be great porting those changes to the OpenShift one, too), `refactor` for the Go files.

MaxFedotov:
> LGTM but I would ask you to split changes in two commits:
> 
> `chore` for the `hack/create-user.sh` (would be great porting those changes to the OpenShift one, too), `refactor` for the Go files.

done

prometherion (approved):

PR #352, opened 2021-07-21
Pod PriorityClass proxy operation support

### LINKED ISSUE #257: Assign Priority Classes to tenants
# Describe the feature
In kubernetes, pods can have priority. Priority indicates the importance of a Pod relative to other Pods. If a Pod cannot be scheduled, the scheduler tries to preempt (evict) lower priority Pods to make scheduling of the pending Pod possible.

In a multi-tenant cluster where not all users are trusted, a tenant owner could create Pods at the highest possible priorities, causing other Pods to be evicted/not get scheduled. 

It would be nice the cluster admin to assign one or more [Priority Classes](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass) to a tenant. The Capsule Operator (i.e. its policy engine component) can control the admitted Priority Classes as for Storage and Ingress Classes.

The cluster admin creates Priority Classes with a given priority number

```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: oil
value: 1000000
globalDefault: false
```

and then assigns it to a tenant

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  priorityClasses:
    allowed:
      - oil
      - default
    allowedRegex: ''
```

# Expected behavior
The tenant admin can only deploy pods with the allowed Priority Classes

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    env: test
spec:
  containers:
  - name: nginx
    image: nginx
    imagePullPolicy: IfNotPresent
  priorityClassName: oil
```

prometherion:
Closes #257, finally.

@MaxFedotov may I ask you if it's ok since you take care of the proxy actions?

### DIFF
--- a/api/v1alpha1/conversion_hub.go
+++ b/api/v1alpha1/conversion_hub.go
@@ -29,33 +29,40 @@ const (
 	ownerUsersAnnotation          = "owners.capsule.clastix.io/user"
 	ownerServiceAccountAnnotation = "owners.capsule.clastix.io/serviceaccount"
 
-	enableNodeListingAnnotation          = "capsule.clastix.io/enable-node-listing"
-	enableNodeUpdateAnnotation           = "capsule.clastix.io/enable-node-update"
-	enableNodeDeletionAnnotation         = "capsule.clastix.io/enable-node-deletion"
-	enableStorageClassListingAnnotation  = "capsule.clastix.io/enable-storageclass-listing"
-	enableStorageClassUpdateAnnotation   = "capsule.clastix.io/enable-storageclass-update"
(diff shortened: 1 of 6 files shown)


MaxFedotov:
lgtm, but can you please also add tests for PriorityClasses here: https://github.com/clastix/capsule/blob/master/api/v1alpha1/conversion_hub_test.go#L129

prometherion:
Thanks a lot @MaxFedotov, there was a bug and writing the test case helped me to find out.

MaxFedotov (approved):
lgtm!

PR #353, opened 2021-07-22
doc: update capsule-proxy docs

viveksyngh:
Signed-off-by: Vivek Singh <vivekkmr45@yahoo.in>


prometherion (approved):
Thanks, definitely a typo we missed!

PR #356, opened 2021-07-22
Tenant or Namespace scope for resource quota budgets

### LINKED ISSUE #50: Make configurable the Resource Quota enforcement at tenant level
# Describe the feature
It would be useful to make configurable via CLI argument, eg. `--force-tenant-quota=true` the resource quota enforcement at tenant level. The default should be true and leave the cluster admin to disable the quota resource enforcement at tenant level.

We would to address case where the cluster admin wants assign resources only at namespace level so the total resource assignment (at tenant level) is statically calculated as number of

`assigned_resources_in_namespace  x namespaceQuota`

For example, with `--force-tenant-quota=false`, the cluster admin can assign 128GB of RAM per namespace and a namespace quota of 3 to a given tenant. So the permitted amount of RAM for that tenant will be 128GB x 3 = 384GB but each namespace has a strict quota of 128GB.

On the opposite, with `--force-tenant-quota=true`, the cluster admin can assign 384GB to the tenant and a namespace quota of 3. So the permitted amount of RAM for a single namespace will be 384GB but this amount of RAM has to be shared between all namespaces.

# Expected behavior
When `--force-tenant-quota=true` (default) we have the current behaviour with creation of ResourceQuota at namespace level and cross namespace (tenant) quota check.

When `--force-tenant-quota=false` the creation of ResourceQuota at namespace level is still in place but no check at cross namespace level (tenant). Only check at namespace level performed by regular kubernetes ResourceQuota admission controller.

prometherion:
Closes #50.

### DIFF
--- a/api/v1alpha1/conversion_hub.go
+++ b/api/v1alpha1/conversion_hub.go
@@ -17,6 +17,8 @@ import (
 )
 
 const (
+	resourceQuotaScopeAnnotation = "capsule.clastix.io/resource-quota-scope"
+
 	podAllowedImagePullPolicyAnnotation = "capsule.clastix.io/allowed-image-pull-policy"
 
 	podPriorityAllowedAnnotation      = "priorityclass.capsule.clastix.io/allowed"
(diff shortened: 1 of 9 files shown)


bsctl:
@prometherion 
Something went wrong: with `spec.resourceQuotas.scope=Tenant`, crossing the pods quota on namespaces, the controller enters in a reconcile loop.

```yaml
  resourceQuotas:
    items:
    - hard:
        limits.cpu: "8"
        limits.memory: 16Gi
        requests.cpu: "8"
        requests.memory: 16Gi
    - hard:
        pods: "10"
        services: "50"
    - hard:
        requests.storage: 10Gi
    scope: Tenant
```

As tenant owner, create:

- two namespaces
- a deployment with replicas 10 in the first namespace
- a deployment with replicas 10 in the second namespace

As cluster admin, check the logs of the controller

prometherion:
What you've just reported is a bug, going to track in a separate issue so we can describe what's going on.

prometherion:
Commit c555206 closes #361.

bsctl (approved):
Lgtm

Issue #148, opened 2020-12-01
Define a benchmark for performances


bsctl:
We should define a benchmark for performances and tests in different stress conditions.

For example:

- how many tenants
- how many namespaces/tenants
- how many concurrent users
- how many requests

prometherion:
Could you elaborate a bit more? Because, at first glance, I'd say it's a matter of Kubernetes limits, here :)

Especially for:

- tenants
- namespaces/tenants
- concurrent users

Regarding requests, I'd say it depends on the resources assigned to Capsule, since we got the webhook server there.

bsctl:
We have to provide a benchmark for adopters. If most of them are Kubernetes limits, let's just to document it.
It would be nice to have a performance test suite, not part of this code base.

prometherion:
I'd say #150 could help here: my plan is to scrape metrics and perform some massive fuzz activities in order to track down the requested reconciliation time, as well as any other metric that can help us define the benchmark.

I'd just say something regarding the maximum amount of resources, as Tenants or Namespaces: this is strictly related to Kubernetes itself, and more specifically, to `etcd`. We're storing the CRDs definitions into it, so we don't have any hard limit since it's something out of our boundaries.

bsctl:
@ptx96 is it possible to use your experience on #300 to provide a benchmark for Capsule performances?

bsctl:
@ptx96 please take care of defining `PrometheusRule` for the Alert Manager. Aims is to inform the cluster admin about possible critical issues about Capsule, for example:

```yaml
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  name: capsule
spec:
  groups:
  - name: capsule-failures
    rules:
    - alert: CapsuleFailures
      annotations:
        message: Capsule reconcile takes too long ....
      expr:
      for: 15m
      labels:
        severity: warning
```

PR #362, opened 2021-07-27
refactor: using goroutines per Namespace for each resource Kind reconciliation

prometherion:
Partially addressing #148 since we're getting a huge performance boost on the expected reconciliation time since we're going to use several goroutines per resource Kind reconciliation (LimitRange, NetworkPolicy, ResourceQuota, etc) and per Namespace.

This will increase the amount of spawned goroutines with subsequent pressure on the memory side: from my actual test, I saw consumption of 1Gi of RAM reconciling a single Tenant collecting 200 Namespaces, not so bad since it's an edge case.

But the most important thing is that we're moving towards to address #49 too: we were doing something really stupid, triggering an update of the outer Resource Quotas per each Namespace, causing several 409 status codes and enqueuing back the Tenant several times. This was leaving enough time frame to force the Tenant Resource Quota budget but since Capsule is now faster, this should be less frequent.

### DIFF
--- a/controllers/tenant_controller.go
+++ b/controllers/tenant_controller.go
@@ -58,8 +58,7 @@ func (r TenantReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
 
 	// Fetch the Tenant instance
 	instance := &capsulev1beta1.Tenant{}
-	err = r.Get(ctx, request.NamespacedName, instance)
-	if err != nil {
+	if err = r.Get(ctx, request.NamespacedName, instance); err != nil {
 		if errors.IsNotFound(err) {
 			r.Log.Info("Request object not found, could have been deleted after reconcile request")
 			return reconcile.Result{}, nil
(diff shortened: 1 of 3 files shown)


controllers/tenant_controller.go
```diff
@@ -292,142 +357,166 @@ func (r *TenantReconciler) syncAdditionalRoleBindings(tenant *capsulev1beta1.Ten
...
-					// sum them and get the Tenant one.
-					rql := &corev1.ResourceQuotaList{}
-					err = r.List(context.TODO(), rql, &client.ListOptions{
-						LabelSelector: labels.NewSelector().Add(*tr).Add(*ir),
-					})
-					if err != nil {
-						r.Log.Error(err, "Cannot list ResourceQuota", "tenantFilter", tr.String(), "indexFilter", ir.String())
-						return err
+			index := i
+			resourceQuota := q
```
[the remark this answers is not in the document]
prometherion: Due to pointers, otherwise, the underlying goroutines started by `group.Go` would refer to the variables provided by the range.

controllers/tenant_controller.go
```diff
@@ -577,6 +636,47 @@ func (r *TenantReconciler) syncNamespaces(tenant *capsulev1beta1.Tenant) (err er
...
+		}
+		// getting NetworkPolicy labels for the mutateFn
+		tl, err := capsulev1beta1.GetTypeLabel(&capsulev1beta1.Tenant{})
+		if err != nil {
+			return err
+		}
+		nl, err := capsulev1beta1.GetTypeLabel(&networkingv1.NetworkPolicy{})
+		if err != nil {
+			return err
+		}
```
MaxFedotov, line 658: This could be moved out of the loop

prometherion: We cannot, we need the index for building the label selector.

controllers/tenant_controller.go
```diff
@@ -292,142 +357,166 @@ func (r *TenantReconciler) syncAdditionalRoleBindings(tenant *capsulev1beta1.Ten
...
-						return err
+			index := i
+			resourceQuota := q
+			g.Go(func() (scopeErr error) {
+				// Calculating the Resource Budget at Tenant scope just if this is put in place.
+				// Requirement to list ResourceQuota of the current Tenant
+				var tr *labels.Requirement
+				if tr, scopeErr = labels.NewRequirement(tenantLabel, selection.Equals, []string{tenant.Name}); scopeErr != nil {
+					r.Log.Error(scopeErr, "Cannot build ResourceQuota Tenant requirement")
+				}
```
MaxFedotov, line 389: this also could be moved out of the loop

controllers/tenant_controller.go
MaxFedotov: i'm talking about these lines
```
		tl, err := capsulev1beta1.GetTypeLabel(&capsulev1beta1.Tenant{})
		if err != nil {
			return err
		}
		nl, err := capsulev1beta1.GetTypeLabel(&networkingv1.NetworkPolicy{})
		if err != nil {
			return err
		}
```
in all other funcs they are out of loop, as label names are static

MaxFedotov:
and one more thing about getting tenant and type labels. 
In some funcs it is like 
```
var tenantLabel string
	if tenantLabel, err = capsulev1beta1.GetTypeLabel(&capsulev1beta1.Tenant{}); err != nil {
		return err
	}
	var typeLabel string
	if typeLabel, err = capsulev1beta1.GetTypeLabel(&corev1.ResourceQuota{}); err != nil {
		return err
	}
```

in others it is 
```
	var tl, ll string
	tl, err = capsulev1beta1.GetTypeLabel(&capsulev1beta1.Tenant{})
	if err != nil {
		return
	}
	ll, err = capsulev1beta1.GetTypeLabel(&rbacv1.RoleBinding{})
	if err != nil {
		return
	}
```
and in some it is 
```
	var tl string
	if tl, err = capsulev1beta1.GetTypeLabel(&capsulev1beta1.Tenant{}); err != nil {
		return
	}
	var ll string
	if ll, err = capsulev1beta1.GetTypeLabel(&corev1.LimitRange{}); err != nil {
		return
	}
```

let's make in constant may be?

controllers/tenant_controller.go
prometherion: No, this cannot be moved out: the `Requirement` is used to combine labels selection, based on the index.

prometherion:
> WDYT about making this more constant?

Done!

MaxFedotov (approved):
lgtm!

PR #375, opened 2021-08-07
refactor: split tenant controller to separate files

### LINKED ISSUE #363: Refactoring the Tenant controller
tenant_controller.go codebase became too large to read, so I propose to split it into separate files for each of the resources it manages, like
```
controllers\tenant\
reconciler.go
utils.go
networkpolicy.go
limitrange.go
resourcequota.go
rolebinding.go
namespace.go
```

MaxFedotov:
closes #363


prometherion (approved):
LGTM, except for the fact the subject should be `refactor` rather than `chore`: going to squash this and amend the message.

PR #389, opened 2021-08-17
Fix PriorityClasses description in CRD

### LINKED ISSUE #387: wrong description for priorityClasses field
# Bug description
The `kubectl explain tenant.spec.priorityClasses` command reports a wrong description

```
Specifies the allowed IngressClasses assigned to the Tenant. Capsule
     assures that all Ingress resources created in the Tenant can use only one
     of the allowed IngressClasses. Optional.
```

Just a typo. It should be

```
Specifies the allowed priorityClasses assigned to the Tenant. Capsule
     assures that all Pod resources created in the Tenant can use only one
     of the allowed PriorityClasses. Optional.
```

bsctl:
Close #387


prometherion (changes requested):
The change must be ported to the Chart, too.

prometherion (approved):

PR #384, opened 2021-08-16
additional servicemonitor helm values support

ptx96:
Close #383

PR #391, opened 2021-08-17
build(helm): add customLabels and customAnnotations params

### LINKED ISSUE #386: support `customLabels` in capsule helm chart
Sometimes users had some custom logic, which relies on labels that should be present on all application parts (deployment, service, rbac, etc). 

In many charts this is done via `customLabels` param in `values` (see https://github.com/kyverno/kyverno/blob/main/charts/kyverno/templates/_helpers.tpl#L40 as an example). 

@prometherion, @bsctl  What do you think about adding this param to `capsule` helm chart?

MaxFedotov:
closes #386

### DIFF
--- a/charts/capsule/README.md
+++ b/charts/capsule/README.md
@@ -94,7 +94,8 @@ Parameter | Description | Default
 `serviceMonitor.enabled` | Specify if a Service Monitor must be created. | `false`
 `serviceMonitor.serviceAccount.name` | Specify Service Account name for metrics scrape. | `capsule`
 `serviceMonitor.serviceAccount.namespace` | Specify Service Account namespace for metrics scrape. | `capsule-system`
-
+`customLabels` | Additional labels which will be added to all resources created by Capsule helm chart . | `{}`
+`customAnnotations` | Additional annotations which will be added to all resources created by Capsule helm chart . | `{}`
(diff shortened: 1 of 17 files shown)


ptx96:
This PR will probably have conflicts with #384 due to changes inside servicemonitor.yaml ;
we have to decide whether to leave servicemonitors' own values or not.

@prometherion WDYT?

prometherion:
You're right, @ptx96: we have to have separate values for the ServiceMonitor since the Prometheus Operator would require some matching labels, and applying these to non-ServiceMonitor resources would be silly.

@MaxFedotov may I ask you this change?

MaxFedotov:
> You're right, @ptx96: we have to have separate values for the ServiceMonitor since the Prometheus Operator would require some matching labels, and applying these to non-ServiceMonitor resources would be silly.
> 
> @MaxFedotov may I ask you this change?

Sorry, don't understand what I had to do :( rebase on #384?

prometherion:
My bad @MaxFedotov :)

Once #384 is merged, rebase over `master`, a git conflict should arise and must be fixed.

MaxFedotov:
@prometherion  rebased on master

prometherion (approved):

PR #392, opened 2021-08-17
Enforcement of LoadBalancer service kind

### LINKED ISSUE #390: Prevent tenants to use ServiceType LoadBalancer
# Describe the feature
In a multi-tenant environment, it would be useful to prevent tenant owners to use ServiceType LoadBalancer. The reason for that is  ServiceType LoadBalancer creates an external Load Balancer cloud resource (when allowed by the Cloud Provider) and this can be not allowed by the cluster admin policy.

To prevent tenant users to allocate external cloud Load Balancers, it would be useful to add to tenant CRD the following option:

```yaml
  serviceOptions:
    allowedServices:
      loadBalancer: false
```
Default value is true.

# Expected behavior
ServiceType LoadBalancer can be not allowed for tenants

prometherion:
Closes #390.

Already updated:
- [x] tests
- [x] kustomize and installer files
- [x] Helm charts


docs/operator/use-cases/service-type.md
```diff
@@ -43,7 +43,29 @@ spec:
 EOF
 ```
 
-With the above configuration, any attempt of Alice to create a a Service of type `externalName` is denied by the Validation Webhook enforcing it. Default value is `true`.
+With the above configuration, any attempt of Alice to create a Service of type `externalName` is denied by the Validation Webhook enforcing it. Default value is `true`.
+
+## LoadBalancer
+
+Same sa previously, Service of type of `LoadBalancer` could be blocked for various reasons. To prevent tenant owners to create these kinds of services, the cluster admin can prevent a tenant to create them:
```
bsctl, line 50: just a typo: Same as previously

bsctl (approved):
LGTM! @prometherion thanks

PR #404, opened 2021-08-20
Change webhook metric used in Grafana dashboard

### LINKED ISSUE #403: Grafana Webhook variable does not provide all possible values until triggered
# Bug description

We currently fetch the "Webhook" variable' values from the _controller_runtime_webhook_latency_seconds_sum_ metric
However, this solution does not provide all possible values **until relative webhooks are triggered**

# How to reproduce

Simply upload the capsule dashboard **without** executing any trigger/test.

# Expected behavior

image

bsctl:
Close #403

### DIFF
--- a/config/grafana/dashboard.json
+++ b/config/grafana/dashboard.json
@@ -14,7 +14,7 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "7.5.5"
+      "version": "7.5.4"
     },
     {
       "type": "panel",
@@ -58,7 +58,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": null,
-  "iteration": 1627970615905,
+  "iteration": 1629463154511,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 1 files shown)


ptx96:
I've uploaded this new dashboard and it seems to work flawlessly.
Now we list also _convert_ webhook, but this one does not appear to be triggered by the current e2e tests.

image

@prometherion WDYT?

prometherion:
> Now we list also _convert_ webhook, but this one does not appear to be triggered by the current e2e tests

That's ok, the `/convert` endpoint will be triggered when dealing with old versioned resources (`v1alpha1`).

You can test manually creating an outdated version (`kubectl apply -f config/samples/capsule_v1alpha1_tenant.yaml`) or forcing the retrieval of `v1beta1` resources using `kubectl get tenants.v1alpha1.capsule.clastix.io oil`.

ptx96 (approved):
Maybe we should also update [monitoring.md](https://github.com/clastix/capsule/blob/master/docs/operator/monitoring.md) inside webhook available values..

anyway, definitely LGTM!

PR #406, opened 2021-08-21
docs: updating capsule-proxy

prometherion:
Porting changes from clastix/capsule-proxy#132.


bsctl (approved):
lgtm @thanks for taking care of it!

PR #417, opened 2021-09-10
Make programmable deny of wildcard hostnames

### LINKED ISSUE #219: Make programmable deny of wildcard hostnames
Let's also assume, that we are using some type of automatization, in order to publish ingress hosts to DNS automatically (like https://github.com/kubernetes-sigs/external-dns/)

As a tenant-owner of `gas`, I create ingress with the host like `- host: "*.bigorg.com"`. That can lead to big problems for an oil tenant, which may create ingress with `-host: oil.bigorg.com` (and for other tenants in a cluster)

But on the other hand, guys from `gas` Tenant can buy a domain like `mygas.com`, which will be used only by them. And they want to create a single ingress in order to handle all requests to `*.mygas.com`. 

So there should be a list of domains for Tenants, where wildcard ingress host may be created. All other wildcard ingress hosts should be denied by a webhook (if it is enabled)

_Originally posted by @MaxFedotov in https://github.com/clastix/capsule/issues/214#issuecomment-790486783_

alegrey91:
Closes #219

### DIFF
--- a/api/v1beta1/deny_wildcard.go
+++ b/api/v1beta1/deny_wildcard.go
@@ -1,0 +1,15 @@
+// Copyright 2020-2021 Clastix Labs
+// SPDX-License-Identifier: Apache-2.0
+
+package v1beta1
+
+const (
+	denyWildcard = "capsule.clastix.io/deny-wildcard"
+)
+
+func (t *Tenant) IsWildcardDenied() bool {
+	if v, ok := t.Annotations[denyWildcard]; ok && v == "true" {
+		return true
+	}
+	return false
+}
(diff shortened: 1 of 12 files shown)


MaxFedotov:
@alegrey91 @prometherion Hey! And why not to add `denyWildcard` to `spec.IngressOptions`, as this would be backward-compatible change?

prometherion (commented):
Looking at changes everything seems good, however we got failing tests on ingressClass check, still investigating why this happens.

prometherion:
> @alegrey91 @prometherion Hey! And why not to add `denyWildcard` to `spec.IngressOptions` instead of using annotations, as this would be backward-compatible change?

That would require a CRD change, requiring a new minor release: idea is to use the annotations, and then prepare the struct change for the new API version.

Does it sound good to you, @MaxFedotov ?

Issue #418, opened 2021-09-13
JQ installation should be checked in hack/create_user.sh


mendrugory:
# Bug description

`hack/create-user.sh` uses [jq](https://stedolan.github.io/jq/) but it has to be installed in the working machine. A previous checking would be good.

# How to reproduce

1. Uninstall `jq` if it is installed
2. Run `hack/create-user.sh` user tenant

# Expected behavior

```bash
$ ./create-user.sh  user tenant
Error: jq not found
```

PR #419, opened 2021-09-13
jq installation checking

mendrugory:
This PR will try to fix the issue described over [here](https://github.com/clastix/capsule/issues/418).
It will add the `jq` installation checking to the `hack/create_user.sh` script


prometherion (changes requested):
Please, amend the commit using the `fix` message type.

prometherion (approved):

PR #423, opened 2021-09-15
Namespace labeling for tenant owners

### LINKED ISSUE #407: support namespace labeling for tenant owners
Current capsule and capsule-proxy implementations deny namespace labeling for tenant owners. The background for this decision was that there are a number of labels, which can lead to different security issues when they are added to a namespace (for example, they can allow to bypass Network Policies).

While this is true, sometimes tenant owners need to have and ability to add some well-known labels defined by their environment. For example, this can be different integrations with 3rd party systems (monitoring, configuration management) where we need to match some business applications with Kubernetes namespaces using some predefined and established set of labels.

So to support these cases and not to compromise overall security I propose adding additional `allowedLabels` list to `namespaceOptions` field in `tenantSpec`, where cluster-administrator can provide a list of labels, that tenant owners can modify via `capsule-proxy`

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - name: alice
    kind: User
  namespaceOptions:
    allowedLabels:
     - myorg.net/foo
     - someother.com/bar
```

This will require an additional webhook for namespaces, where we allow modification of only allowed labels for a namespace.

@bsctl, @prometherion WDYT?

MaxFedotov:
closes #407 
@prometherion will add documentation tomorrow :)


MaxFedotov:
seems like lint error is related to https://github.com/conventional-changelog/commitlint/issues/2141
@prometherion  can you please take a look?

config/install.yaml
```diff
@@ -1411,7 +1451,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        image: quay.io/clastix/capsule:v0.1.0
+        image: quay.io/clastix/capsule:v0.1.0-rc5
```
prometherion, line 1454: Please, perform a `git fetch --all` on your local repository to ensure downloading all the git tags since this is going to be updated by the makefile using the latest git tag.

prometherion (changes requested):

config/install.yaml
MaxFedotov: done

MaxFedotov:
@prometherion oh yes, all green :) waiting for a review!

MaxFedotov:
rebased on a fresh master

charts/capsule/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -32,6 +32,7 @@ webhooks:
       - v1
       operations:
       - CREATE
+      - UPDATE
```
prometherion, line 35: Changes on Chart requires a version bump up in order to build a new release.

prometherion (changes requested):
LGTM, except the missing on Helm Chart version bump up.

charts/capsule/templates/mutatingwebhookconfiguration.yaml
MaxFedotov: should i also update `appVersion` in chart?

prometherion: Since we don't have a new image, nor a release candidate, we can skip the chart bump.

prometherion (approved):

PR #425, opened 2021-09-19
fix doc minor issues

brightzheng100:
IMHO, this is the best set of docs regarding what considerations must be taken while thinking/designing multitenancy support in Kubernetes.

While walking the docs through one by one, I spotted some small issues and this is a small PR to fix all I found.

### DIFF
--- a/docs/operator/use-cases/create-namespaces.md
+++ b/docs/operator/use-cases/create-namespaces.md
@@ -36,20 +36,20 @@ subjects:
   name: alice
 roleRef:
   kind: ClusterRole
-  name: namespace-deleter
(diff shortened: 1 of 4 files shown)


bsctl (approved):
@brightzheng100 Thank you very much for submitting this PR. Really appreciated.

bsctl:
@brightzheng100 I forget to mention we use [conventional commit messages](https://github.com/conventional-changelog/commitlint/#what-is-commitlint). Please change the commit message to something like `docs: fix doc minor issues`. Thanks

brightzheng100:
Ah...that makes sense and it's been updated. Thanks

PR #433, opened 2021-09-23
add documentation about deny-wildcard annotation

### LINKED ISSUE #219: Make programmable deny of wildcard hostnames

alegrey91:
fixes #219.

### DIFF
--- a/docs/operator/use-cases/deny-wildcard-hostnames.md
+++ b/docs/operator/use-cases/deny-wildcard-hostnames.md
@@ -1,0 +1,32 @@
+# Deny Wildcard Hostnames
+
+Bill, the cluster admin, can deny the use of wildcard hostnames.
+
+Let's assume that we had a big organization, having a domain `bigorg.com` and there are two tenants, `gas` and `oil`.
+
+As a tenant-owner of `gas`, Alice create ingress with the host like `- host: "*.bigorg.com"`. That can lead to big problems for the `oil` tenant because Alice can deliberately create ingress with host: `oil.bigorg.com`.
+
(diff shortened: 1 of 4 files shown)


prometherion (changes requested):
LGTM, just needs to be rebased over the latest `master`.

prometherion (approved):

aminmr:
I have a specific domain for my customers, such as freeexample.com, which is available for general use. However, for certain special users, I need to restrict access to creating an ingress with this domain. Do you have any ideas on how to achieve this?
@prometherion

prometherion:
@aminmr I would suggest you use the [Slack workspace channel](https://kubernetes.slack.com/archives/C03GETTJQRL) to ask for more information about the Capsule features.

PR #463, opened 2021-10-26
feat(helm): find kubectl tag from server version

### LINKED ISSUE #462: Find helm job kubectl tag from cluster version
# Describe the feature

When installing / uninstalling capsules via helm, helm jobs make use of a kubectl client.
This client should have a version that is as compatible as possible with the target kubernetes cluster.

1. Start a capsule installation with helm ;
2. Post-install job -> helm dynamically chooses which kubectl tag is best to pull down based on the cluster version ;
3. Uninstall capsule ;
4. Post-delete job -> helm dynamically chooses which kubectl tag is best to pull down based on the cluster version.

# Expected behavior
Helm dynamically chooses which kubectl tag is best to pull down based on the cluster version

ptx96:
Closes #462

if applied, this commit removes the static kubectl client tag and will use a dynamic tag based on the cluster version.

- it'll use the last tag of the matching "minor"
- it is always possible to override the value

this PR remains temporarily in draft, waiting for changes to the push logic of clastix/kubectl repository

### DIFF
--- a/charts/capsule/templates/_helpers.tpl
+++ b/charts/capsule/templates/_helpers.tpl
@@ -95,7 +95,12 @@ Create the proxy fully-qualified Docker image to use
 Create the jobs fully-qualified Docker image to use
 */}}
 {{- define "capsule.jobsFullyQualifiedDockerImage" -}}
+{{- if .Values.jobs.image.tag }}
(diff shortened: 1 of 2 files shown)


prometherion:
@ptx96 clastix/kubectl#8 is approved, please, proceed.

prometherion (approved):
Wonderful, LGTM!

PR #464, opened 2021-10-26
Forbidden node labels and annotations

### LINKED ISSUE #435: Forbidden node labels and annotations
Currently, we can allow tenant admins to modify nodes using `capsule-proxy` if appropriate permission is set in `tenant CRD`.

While this is a great option for them, which adds additional flexibility, there are some types of labels or annotations, which must be protected from modifications (for example, which are set by `cloud-providers` or `autoscalers`)

My proposal is to add an ability for cluster administrators to protect some labels or annotations from modification by tenant owners (the same, as it is done with namespaces)

If the `cluster administrator` wants to disallow `tenant owner` to add some labels or annotations on nodes, he had to add one of the following annotations on a tenant:

- capsule.clastix.io/forbidden-nodes-labels
- capsule.clastix.io/forbidden-nodes-labels-regexp
- capsule.clastix.io/forbidden-nodes-annotations
- capsule.clastix.io/forbidden-nodes-annotations-regexp

When new capsule api will be release, these annotations will be moved to the new `NodeOptions` struct (together with `nodeSelector`:

```go
type NodeOptions struct {
        // Specifies the label to control the placement of pods on a given pool of worker nodes. All namespaces created within the Tenant will have the node selector annotation. This annotation tells the Kubernetes scheduler to place pods on the nodes having the selector label. Optional.
	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
	// Specifies forbidden a labels the Tenant owner can't place on Nodes resources in the Tenant. Optional.
	ForbiddenLabels *ForbiddenListSpec `json:"forbiddenLabels,omitempty"`
	// Specifies forbidden annotations the Tenant owner can't place on Nodes resources in the Tenant. Optional.
	ForbiddenAnnotations *ForbiddenListSpec `json:"forbiddenAnnotations,omitempty"`
}
```

@bsctl @prometherion WDYT?

MaxFedotov:
closes #435


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
prometherion, line 1: Why do we have this? No update of `controller-gen` dependency, wondering why CI is not complaining, too.

docs/operator/use-cases/node-labels-and-annotations.md
```diff
@@ -0,0 +1,40 @@
+# Denying specific user-defined labels or annotations on Nodes
+
+**Important note**
```
prometherion, line 3: Maybe we could use the blockquotes for this important now, and maybe at the end of the section?

prometherion (approved):
Just a minor concern on autogenerated files, overall LGTM!

MaxFedotov:
@prometherion i was on vacations, will fix everything you mentioned and merge it today in the evening

api/v1alpha1/zz_generated.deepcopy.go
MaxFedotov: seems like something with my local `controller-gen`. I've removed this changes

docs/operator/use-cases/node-labels-and-annotations.md
MaxFedotov: done :)

Issue #486, opened 2021-12-20
Helm chart - imagePullSecrets are not applied to post-install-job and pre-delete-job


slushysnowman:
# Bug description

To prevent external dependencies, we are replicating Capsule images to a private repo, for which we need to auth. However, we run into issues with the post-install-job and pre-delete jobs, as these don't have imagePullSecrets specified in the spec.

- https://github.com/clastix/capsule/blob/master/charts/capsule/templates/post-install-job.yaml
- https://github.com/clastix/capsule/blob/master/charts/capsule/templates/pre-delete-job.yaml

# How to reproduce

1. Store kubectl and capsule images in a private repo
2. Specify imagePullSecrets in helm chart values and modify image source repository to private repo
3. Install helm chart using values

# Expected behavior

imagePullSecrets specified via helm chart values should also be applied to post-install-job and pre-delete-job if specified.

# Logs

If applicable, please provide logs of `capsule`.

In a standard stand-alone installation of Capsule,
you'd get this by running `kubectl -n capsule-system logs deploy/capsule-controller-manager`.

# Additional context

- Capsule version: 0.1.0
- Helm Chart version: v0.1.3

prometherion:
Hey @slushysnowman, you're totally right: would you like to contribute to this?

Just let me know so I can track it properly!

slushysnowman:
Yep, I'll work on this, was literally about to start.

prometherion:
[Chart has been released](https://github.com/clastix/charts/blob/gh-pages/capsule-0.1.4.tgz), update your local repository! 👀

PR #487, opened 2021-12-20
fix: add imagePullSecrets to jobs

slushysnowman:
Add imagePullSecrets to post-install-jobs and pre-delete-jobs
Closes https://github.com/clastix/capsule/issues/486

### DIFF
--- a/charts/capsule/templates/post-install-job.yaml
+++ b/charts/capsule/templates/post-install-job.yaml
@@ -25,6 +25,10 @@ spec:
         app.kubernetes.io/instance: {{ .Release.Name | quote }}
         helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       restartPolicy: Never
       containers:
       - name: post-install-job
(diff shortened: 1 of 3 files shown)


charts/capsule/templates/post-install-job.yaml
```diff
@@ -25,6 +25,10 @@ spec:
         app.kubernetes.io/instance: {{ .Release.Name | quote }}
         helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     spec:
+      {{- with .Values.imagePullSecrets }}
```
prometherion, line 28: Should we use the Manager pull secrets here, or maybe, would be better to provide different secrets just for this Job, such as `jobs.image.imagePullSecrets`?

slushysnowman: I'm doing a lot of stuff with charts at the moment - most places I see it just defined at global level for all images in the chart. Other charts like [bitnami/external-dns](https://github.com/bitnami/charts/blob/master/bitnami/external-dns/values.yaml) use a global value as well as individual values.

Either works for me to be honest - I'd be tempted to just use the main imagePullSecrets for simplicity's sake, and if a use case comes about where it's better to specify it separately for the jobs, then add it then, but if you think it's better to separate it from the get-go just let me know and I'll do that.

prometherion: No problem, it was just to start the discussion.

Let's implement this way, we have always room for improvement :)

prometherion (approved):
LGTM, if you want to get a patch release of this Chart, just bump up the version in the `Chart.yaml` file so we can tag it!

slushysnowman:
@prometherion - Chart version bupmped and ready for merge! Thanks for the review!

PR #539, opened 2022-04-06
fix(helm): jobs in capsule helm chart should use the same tolerations as deployment

### LINKED ISSUE #538: post and pre jobs in capsule helm chart should use the same tolerations as capsule deployment
When installing capsule from helm chart and adding `tolerations` values (e.g  effect: NoSchedule, key: node-role.kubernetes.io/master) capsule installation will fail because the `post-install-job` will not be able to start on master nodes due to missing tolerations in job spec

MaxFedotov:
closes #538


prometherion (approved):
LGTM

PR #542, opened 2022-04-11
fix Makefile for building local binary with make manager

gernest:
I tried to run `make manager` to make sure I can build capsure locally.

-> First attempt

```
make: *** No rule to make target `fmt', needed by `manager'.  Stop.
```

fmt and vet rules were missing. I added fmt and vet rules(as part of this commit)

-> Second attempt

```
go fmt ./...
go vet ./...
go build -o bin/manager main.go
./main.go:65:55: undefined: GitTag
./main.go:65:63: undefined: GitCommit
./main.go:65:74: undefined: GitDirty
./main.go:66:46: undefined: GitRepo
./main.go:67:46: undefined: BuildTime
make: *** [manager] Error 2
```

This is due to  passing main.go to build, while the package has multiple files
`main.go` and `version.go`, the missing symbols are in `version.go`

I fixed this by building root package. And now go is happy!

-> Third attempt

```
go fmt ./...
go vet ./...
go build -o bin/manager
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,8 @@ test: generate manifests
 	go test ./... -coverprofile cover.out
 
 # Build manager binary
-manager: generate fmt vet
-	go build -o bin/manager main.go
+manager: generate golint
+	go build -o bin/manager
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate manifests


Makefile
```diff
@@ -41,12 +41,20 @@ test: generate manifests
 
 # Build manager binary
 manager: generate fmt vet
```
prometherion, line 43: We already have linting enabled, you can use the target `golint` that is containing all the required rules.

Makefile
```diff
@@ -41,12 +41,20 @@ test: generate manifests
...
 	go run .
 
+# Run go fmt against code
+fmt:
+	go fmt ./...
+
+# Run go vet against code
+vet:
+	go vet ./...
+
```
prometherion, line 57: No need to import back `fmt` and `vet`, let's rely only on target `golint`.

prometherion (changes requested):
Thanks @gernest for the hotfix, really appreciated it! 

Could you address the required changes? Once done, LGTM!

gernest:
@prometherion I made the changes PTAL.

gernest:
Done!

Thanks.

prometherion (approved):

PR #545, opened 2022-04-12
feat: expose custom metrics for tenants status

### LINKED ISSUE #451: Tenant Metrics
# Describe the feature

We would like to have more metrics being exported about the current tenant controlled by an operator. Some Metrics t hat would be helpful:

- **Basic info metric** (Tenant Active State, Namespace Quota and used Namespaces, so we can count up how many tenants there are)
- **Cordoned tenants** (Consider the tenant cordoning label, to tell which tenants are cordoned and which are not)
- **Quota Usage** (Evaluate how much of o quota spanned over a tenant is used vs max given quota). Same on namespace basis

These are the most important ones i could think of. Another interesting feature (which most other metric exporter are lacking) ist the possibility to add labels to my resource which then are added as metric label. So le't's say I want to be able to show all tenants of certain customer. I would add the label `metrics.clastix.io/customer=a` to a tenant cr and then the label `customer`with the value `a` would show up in the metric. This way everyone has much greater flexibility to organize metrics, even if they come from the same controller. Simply if you would check if there's any label matching `metrics.clastix.io/*` and then register it with the actual metric.

Doesn't change.

# Expected behavior
A clear and concise description of what you expect to happen.

gernest:
closes #451

This commit adds pkg/stats that defines prometheus counters for  active and
cordoned tenants.

These metrics are exposed to controller-runtime. So they are visible on the
controller metrics endpoint.

This was tested manually , after creating a new tenant the metrics showed
up on the metrics service endpoint

```
# HELP capsule_tenants_status_active_count Total number of active tenants
# TYPE capsule_tenants_status_active_count counter
capsule_tenants_status_active_count 1
```

### DIFF
--- a/controllers/tenant/manager.go
+++ b/controllers/tenant/manager.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	capsulev1beta1 "github.com/clastix/capsule/api/v1beta1"
+	"github.com/clastix/capsule/pkg/stats"
 )
 
 type Manager struct {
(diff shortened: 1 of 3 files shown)


adberger:
This is a good start :)  
@prometherion any updates on this?

MaxFedotov:
what do you think about using https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md? 
It allows to add custom metrics based on fields in CRD without a need to modify code

adberger:
> what do you think about using https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md? It allows to add custom metrics based on fields in CRD without a need to modify code

Solved according to your suggestion, made an example in the referenced issue (https://github.com/projectcapsule/capsule/issues/451).

prometherion:
Closing in favour of https://github.com/projectcapsule/capsule/issues/451#issuecomment-1774537482.

Issue #525, opened 2022-03-11
Propagate resources across all Tenant Namespace using the TenantResource custom resource


prometherion:
# Current design

Alice would like to create several resources on each of their Namespaces. Upon the creation of each Namespace, they have to create the desired resources on each Namespace. Any automation cannot be put in place here.

# Desired design

A new CRD could be crafted as follow.

```yaml
apiVersion: capsule.clastix.io/v1beta2
kind: TenantResource
metadata:
  name: database
  namespace: solar-management
spec:
  resyncPeriod: 60s
  additionalResources:
    namespaceSelector:
      matchLabels:
        tier: one
    items:
    - apiVersion: presslabs.io/v1beta1
      kind: MySQL
      spec:
        foo: bar
  clusterRoles:
    namespaceSelector:
      matchLabels:
        tier: one
    items:
      - name: database-admin
        subjets:
        - kind: Group
          name: developers
```

The Namespace-scoped resource `TenantResource` would allow `alice` to create on each Tenant namespace resource declared in the `/spec/additionalResources` key.

These resources (like a DB CRD) would require RBAC to `developers`, granted using the `/spec/clusterRoles` specification key.

Replication of these resources could be furthermore fine-grained using a `namespaceSelector`, allowing to define on which Namespace these should be replicated.

A reconciliation to ensure the selected status is expected is controlled by the `resyncPeriod`, a time-based value that allows to poll the current state of the resources and ensure the desired one.

This new custom resource definition is going to be defined as Namespace-scoped in order to allow Tenant owners to create their resources in a fashion automated way.

# Proposal contributors

- @MaxFedotov 
- @prometherion 
- @carpenterm

Looking for contributors for this feature!

slushysnowman:
Looks pretty good - would this also allow specifying pre-existing resources to be created in namespaces? Or would the spec of the item to be created in the namespaces have to be defined within `TenantResource`?

I'm imagining this could also be used to solve this issue? https://github.com/clastix/capsule/issues/222 - my only concern is if we specify secrets on this spec (although maybe we can use a CRD to retrieve them from elsewhere)

oliverbaehler:
Sure thing, that's a great proposal and would certainly help with #416.

For the selector it would be awesome if we could also use `matchExpressions` instead of only `namespaceSelector`. This would allow more generic assignments, eg:

```
apiVersion: capsule.clastix.io/v1beta2
kind: TenantResource
metadata:
  name: database
  namespace: solar-management
spec:
[...]

Issue #416, opened 2021-09-09
Support Any Additional resource


oliverbaehler:
# Describe the feature

I would like to have a generic approach on resources, which are created for each new namespace belonging to a tenant. The same behavior that `.spec.additionalRoleBindings` and `.spec.networkPolicies` already implement, but for any kubernetes resource.

As of why:

We are currently moving to Cilium Network Policies, and we would require each namespace to implement those policies (same could also apply for calico resources or really anything else). So instead of writing an own workaround, it would make sense to be able to declare something like this:

```
---
apiVersion: capsule.clastix.io/v1beta1
kind: Tenant
metadata:
  name: gas
spec:
  additionalResources:
    -  apiVersion: "cilium.io/v2"
       kind: CiliumNetworkPolicy 
       metadata:
         name: "l3-rule"
      spec:
        endpointSelector:
          matchLabels:
             role: backend
       ingress:
       - fromEndpoints:
         - matchLabels:
             role: frontend
```

# What would the new user story look like?

How would the new interaction with Capsule look like? E.g.

1. A new Tenant is created with additional resources
2. The resources could be validated on tenant creation and reject the tenant creation if there's an error, if not the tenant is created.
3. A new User creates/assigns a new namespace to the previously created tenant.
4. the additional resources are added to the namespace once the operator has reconciled.
5. Everyone is happy.

# Expected behavior
A clear and concise description of what you expect to happen.

When I have defined `.spec.additionalResources` on a tenant, those resources are created for each namespace that is assigned or created in that tenant.

bsctl:
@oliverbaehler thanks for suggesting this improvement. I think this will require some sort of refactoring of the code. Let's to see what @prometherion says.

prometherion:
I'm a bit late on this, thanks for the patience!

This feature would be awesome and I couldn't agree more, despite some limitations I'd like to evaluate with the community.

We have to keep in mind that CRDs are, essentially, OpenAPI v3 specifications: we got a schema that must stick to it and that's great, I love it when everything is strictly typed since it's less error-prone and safer, especially at compile time.
The first challenge would be how to structure the `/spec/additionalResources` field: although from a Go struct type point of view a `[]interface{}` type could do the trick, the resulting OpenAPI v3 wouldn't be so happy (ndr: not sure, should be tested), and even if it wouldn't be the case, we would have to deal with another issue, how to apply these resources.

[...]

PR #548, opened 2022-04-18
feat: support additional resources across tenant namespaces

gernest:
This attempts to address  #525 which is related to #416.

problem this commit is trying to solve
----

```
Alice would like to create several resources on each of their Namespaces.
Upon the creation of each Namespace, they have to create the desired resources on each Namespace
```

On the above linked tickets there are two proposed approaches

approach 01
----

Create a new resource `TenantResource`  something like this

```yaml
kind: TenantResource
metadata:
  name: database
  namespace: solar-management
spec:
  resyncPeriod: 60s
  additionalResources:
    namespaceSelector:
      matchLabels:
        tier: one
    items:
    - apiVersion: presslabs.io/v1beta1
      kind: MySQL
      spec:
        foo: bar
  clusterRoles:
    namespaceSelector:
      matchLabels:
        tier: one
    items:
      - name: database-admin
        subjects:
        - kind: Group
          name: developers
```

approach 02
-----

Extend `Tenant` to  support addtional resources

```yaml
kind: Tenant
metadata:
  name: gas
spec:
  additionalResources:
    -  apiVersion: "cilium.io/v2"
       kind: CiliumNetworkPolicy
       metadata:
         name: "l3-rule"
      spec:
        endpointSelector:
          matchLabels:
             role: backend
       ingress:
       - fromEndpoints:
         - matchLabels:
             role: frontend
```

This commit implements approach `02` due to the following reasons

- The namespaces belong to the tenant already, extra `TenantResource` seems redundant given that the lifecycle of the additional resources are tied to the `Tenant`.

How does the crd look like now ?
----

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  resyncPeriod: 60s
  additionalResources:
    namespaceSelector:
      matchLabels:
        tier: one
    items:
      - |
        apiVersion: v1
        kind: Pod
        metadata:
          name: nginx
          labels:
            app.kubernetes.io/name: proxy
        spec:
          containers:
            - name: nginx
              image: nginx:11.14.2
              ports:
                - containerPort: 80
                  name: http-web-svc
      - |
        apiVersion: v1
        kind: Service
        metadata:
          name: nginx-service
          labels:
            app.kubernetes.io/name: proxy
        spec:
          selector:
            app.kubernetes.io/name: proxy
          ports:
            - name: name-of-service-port
              protocol: TCP
              port: 80
              targetPort: http-web-svc
  owners:
    - name: alice
      kind: User
```

The difference with the proposed crd is, items are strings of k8s objects in yaml
format. I ran through decoding issues when I tried to use `Unstructured`  so I
decided to settle on strings instead.

How it works ?
----

We search for namespaces  specified by `namespaceSelector` that are owned by the
tenant. For each matched namespace, apply all resources specified in `additionalResources.items`
on any error, reschedule next reconciliation by `resyncPeriod`.

What is missing ?
----

- [ ] Tests
- [ ] What happens when a tenant is deleted ?
- [ ] Does `additionalRoleBindings` cover for  `clusterRoles` defined in approach 01?

I will wait for feedback/discussion on how to proceed from here.


bsctl:
@gernest @prometherion I'm proposing to close this since no progresses.

TinySong:
@bsctl @gernest this feature is very useful. Why no progresses？

prometherion:
My bad for not providing useful insights on this feature which is highly requested, and I'll try to elaborate properly.

Any new change in the API definition of a CRD, in our case the Tenant one, must follow a specific path for many reasons, although all it could be summed up by the fact that our preferred way to manage the lifecycle of Capsule is using Helm.

A new specification field in the Tenant definition would require a change in the `charts/capsule/crds` folder which is containing the Custom Resource manifests, and due to the opinionated way Helm is managing CRDs, these manifests are just installed when the release is missing in the cluster, requiring the Cluster Administrator to patch the `CustomResourceDefinitions` manually. No objection at all to this workflow, it's well known, and it has been deliberately put in place for many reasons I cannot share here since it would be off-topic.

Said so, we tend to implement features using the annotations of the custom resource, and it's really annoying due to the lack of type validation, issues with the max length of the key (RFC DNS-1123), and many other problems.

As maintainers, our idea is to propose this new feature for the upcoming [milestone](https://github.com/clastix/capsule/milestone/4) and we're still arguing on how to implement it properly, by asking for feedback from the community, and talking directly with the adopters, and as here on GitHub, as you could have seen.

I absolutely value the @gernest proactivity in trying to address this feature on his own, just a bit awkward fact that most of the hard work must be revisited, and since this is going to be a heavy feature, it should be managed by the maintainers itself to ensure a reliable implementation, and avoid eventually SPOF with the adopters that are contributing during their free time.

The main issue here has been my inactivity in pointing out all of these critic points before getting this PR opened, and a bit awkward that all the effort showed in this contribution will be lost, although we, as a community, are always welcoming contributions, and we're actively looking forward to it: syncing people, in a distributed and asynchronous world, is really challenging, I'll try to take note and behave more wisely in the future.

PR #553, opened 2022-05-02
Code base improvement

prometherion:
Not tracked with an issue, just enabling the linters and fixing minor stuff.


controllers/servicelabels/abstract.go
```diff
@@ -8,16 +8,16 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
```
MaxFedotov, line 13: shouldn't it be imported as:
`apierrors "k8s.io/apimachinery/pkg/api/errors"`

MaxFedotov (approved):
lgtm, except one comment :)

controllers/servicelabels/abstract.go
arriqaaq: yes, apierrors has been used across

PR #555, opened 2022-05-06
support cert-manager for generating tls and ca

### LINKED ISSUE #554: Use cert-manager to issue certificate for the capsule webhooks
# Describe the feature

the generation of the `capsule-tls` and `capsule-ca,` we are doing this through built-in controllers in the capsule and trying to 
remove any other dependency required for the capsule to run.

But can we make this feature somehow optional and provide another way to provide the tls cert and ca cert? For other operators, we use `cert-manager` to inject ca and tls certs for webhooks, we would like to use the same thing for the capsule as well. If we can make these controllers optional and the certs path configurable then we can use `cert-manager` to inject certificates.

https://github.com/clastix/capsule/issues/439#issuecomment-1082589467

What would the new interaction with Capsule look like? E.g.

1. What are the prerequisites for this?
4. All the magic happens in the background

Feel free to add a diagram if that helps explain things.

# Expected behavior
A clear and concise description of what you expect to happen.

MaxFedotov:
closes #554 

Added new flag to Capsule `enable-secret-controller`, which enables or disables secretcontroller reconcilers.
added 2 new options to Capsule helm chart:
- `manager.options.enableSecretController` - configures `enable-secret-controller` flag
- `certManager.generateCertificates` - enables cert-manager certificates yaml and other stuff

### DIFF
--- a/main.go
+++ b/main.go
@@ -70,13 +70,15 @@ func printVersion() {
 
 // nolint:maintidx
 func main() {
-	var enableLeaderElection, version bool
+	var enableLeaderElection, enableSecretController, version bool
 
 	var metricsAddr, namespace, configurationName string
 
 	var goFlagSet goflag.FlagSet
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.BoolVar(&enableSecretController, "enable-secret-controller", true,
+		"Enable secret controller which reconciles TLS and CA secrets for capsule webhooks.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -70,13 +70,15 @@ func printVersion() {
...
 func main() {
-	var enableLeaderElection, version bool
+	var enableLeaderElection, enableSecretController, version bool
 
 	var metricsAddr, namespace, configurationName string
 
 	var goFlagSet goflag.FlagSet
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.BoolVar(&enableSecretController, "enable-secret-controller", true,
```
prometherion, line 80: Do we need this flag?

If we're missing the Secret since non-existing, the Pod will not be started until those are available for the mount.

MaxFedotov: we need to disable build-in reconciler to be able to generate secret using cert-manager

prometherion: I don't have any doubts about it, pretty sure it works. I'm just wondering why we should disable the secret reconciler since the CA and TLS would be created by `cert-manager`.

Capsule would just check if the Secret CA and TLS are there, and read their value. Pretty sure I'm missing something.

MaxFedotov: due to this https://github.com/clastix/capsule/blob/master/controllers/secret/ca.go#L183. The controller will find a secret, but it will try to update it with its own generated data, which will lead to 
```
2022/05/06 11:50:35 http: TLS handshake error from 10.244.0.1:1547: remote error: tls: bad certificate
```
errors

MaxFedotov: and there is one more problem. I recalled that we need to inject cert-manager annotation to tenant crd, as we also have a conversion webhook there. And i really don't know how to do it, as crds are not templated

prometherion: @MaxFedotov can't we use the `cert-manager`'s [CA Injector](https://cert-manager.io/docs/concepts/ca-injector/)?

MaxFedotov: We can, but we need to template the name of the certificate in cert-manager annotation '
cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "capsule.fullname" . }}-webhook-cert'

And templates and variables can't be used in crds :(

prometherion: @MaxFedotov WDYT on moving CRD definition from `crds` folder to `templates`? Do you see any drawbacks to this?

We're already minor bumping upon any new CRD update, since we're informing users I don't see any issue on this.

prometherion:
Besides the ongoing discussion, I think we should change the logic here:

https://github.com/clastix/capsule/blob/9f6883d309368a9dc0dda2c79a436837dc924c55/controllers/secret/ca.go#L178-L182

Actually, the code is not aware we have to ignore the secret since managed externally, maybe I could address this in a different issue: WDYT @MaxFedotov ?

main.go
MaxFedotov: The only problem will be that resources in crds folder are not deleted with helm uninstall. If we move them to templates folder, they will be removed (and with the removing of crds all tenants and namespaces will be deleted as well). And one more thing to check is the ordering when applying resources. We have a  capsuleConfiguration crd and capsuleConfiguration resource with default configuration, so we need to ensure that crd will be applied first

MaxFedotov:
> Besides the ongoing discussion, I think we should change the logic here:
> 
> https://github.com/clastix/capsule/blob/9f6883d309368a9dc0dda2c79a436837dc924c55/controllers/secret/ca.go#L178-L182
> 
> Actually, the code is not aware we have to ignore the secret since managed externally, maybe I could address this in a different issue: WDYT @MaxFedotov ?

Completely ok :) you make a separate PR or add a commit to this?

prometherion:
> Completely ok :) you make a separate PR or add a commit to this?

@MaxFedotov we can fix it later, let's get this merged.

prometherion (approved):

PR #556, opened 2022-05-09
fix: build failed  from source on macos

TinySong:


prometherion:
Hey @TinySong, thanks for opening your first PR for Capsule!

Could you elaborate a bit more what's the issue you faced? Actually, we got CI and automated tests and expect Capsule to run on Linux-based machines.

pkg/webhook/tenant/custom_resource_quota.go
```diff
@@ -77,7 +77,7 @@ func (r *resourceCounterHandler) OnCreate(clt client.Client, decoder *admission.
 			}
 
 			if limit, retryErr = capsulev1beta1.GetLimitResourceFromTenant(*tnt, kgv); retryErr != nil {
-				return nil
+				return retryErr
```
prometherion, line 80: This is definitely a bug, returning the error from the upperscope.

Wondering how this is blocking the build as reported in the PR title.

TinySong: thank you reply so quickly. I just recently finding  a multi-tenant solution in Kubernetes. then I find capsule, I think capsule  is Very accord with my expectations. then Clone the code, and execute `make` command.  output the error
 mesage:
```
~/development/golang/src/capsule/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
golangci-lint run -c .golangci.yml
pkg/webhook/tenant/custom_resource_quota.go:80:5: error is not nil (line 79) but it returns nil (nilerr)
                                return nil
                                ^
controllers/servicelabels/abstract.go:31:2: `scheme` is unused (structcheck)
        scheme *runtime.Scheme
        ^
make: *** [golint] Error 1

```

TinySong: maybe I should remove `macOS` in the title of the PR?

prometherion: Yes, please, although is very weird since we got automation.

Anyway, great finding! 👍🏻

TinySong: remove `on macOS ` in the git message.

prometherion (approved):

PR #558, opened 2022-05-09
docs(proxy): documenting retrieval of a single namespace

prometherion:
Documenting new feature introduced with clastix/capsule-proxy#203.


bsctl (approved):
LGTM @prometherion thanks for taking care of that

PR #565, opened 2022-05-16
fix: using sentinel error for non limited custom resource

### LINKED ISSUE #564: Not returning an error in case of non limited custom resource
Actually, we're not discriminating upon the error of parsing the limit for custom resources.

https://github.com/clastix/capsule/blob/8c0c8c653d473ae946b46c1b10c3a1fee10f65ae/pkg/webhook/tenant/custom_resource_quota.go#L79-L81

The function `capsulev1beta1.GetLimitResourceFromTenant` can return an error due to parsing, or because the resource is not limited, and we should do proper error validation.

https://github.com/clastix/capsule/blob/8c0c8c653d473ae946b46c1b10c3a1fee10f65ae/api/v1beta1/custom_resource_quota.go#L36-L39

prometherion:
Closes #564.

### DIFF
--- a/api/v1beta1/custom_resource_quota.go
+++ b/api/v1beta1/custom_resource_quota.go
@@ -32,10 +32,22 @@ func GetUsedResourceFromTenant(tenant Tenant, kindGroup string) (int64, error) {
 	return used, nil
 }
 
+type NonLimitedResourceError struct {
+	kindGroup string
+}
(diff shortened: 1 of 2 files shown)


alegrey91 (commented):
lgtm

alegrey91 (approved):

PR #566, opened 2022-05-16
feat: protected tenant annotation

### LINKED ISSUE #563: Annotation to protect tenant from accidental deletion
Now if we delete a tenant all tenant namespaces are deleted automatically by kubernetes garbage collector. To protect critical tenants from accidental deletion (e.g. wrong commit to ci\cd system) it would be good to have some annotation, like 
`capsule.clastix.io/protected`, which will make `capsule-controller` to ignore tenant deletion and instead add event and message to a log.

@prometherion WDYT?

MaxFedotov:
closes #563

### DIFF
--- a/pkg/webhook/tenant/protected.go
+++ b/pkg/webhook/tenant/protected.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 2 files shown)


MaxFedotov:
@prometherion can you please take a look at tests? don't know why they are failing randomly

pkg/webhook/tenant/protected.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil
+	}
+}
+
+func (h *protectedHandler) OnDelete(_ client.Client, decoder *admission.Decoder, _ record.EventRecorder) capsulewebhook.Func {
+	return func(ctx context.Context, req admission.Request) *admission.Response {
+		tenant := &capsulev1beta1.Tenant{}
+		if err := decoder.Decode(req, tenant); err != nil {
+			return utils.ErroredResponse(err)
+		}
```
prometherion, line 36: The delete request doesn't send any payload, and the resource name can be retrieved from `req.AdmissionRequest.Name`.

It means you have to retrieve it from the local cache, check if it contains the annotation, and then perform your logic.

prometherion (changes requested):

pkg/webhook/tenant/protected.go
MaxFedotov: thanks, fixed!

PR #575, opened 2022-05-30
Switching over Docker Hub image

### LINKED ISSUE #574: Switch default container registry to Docker Hub rather than quay.io

prometherion:
Closes #574.

### DIFF
--- a/charts/capsule/values.yaml
+++ b/charts/capsule/values.yaml
@@ -4,7 +4,7 @@
 
 manager:
   image:
-    repository: quay.io/clastix/capsule
+    repository: clastix/capsule
     pullPolicy: IfNotPresent
     tag: ''
 
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 VERSION ?= $$(git describe --abbrev=0 --tags --match "v*")
 
 # Default bundle image tag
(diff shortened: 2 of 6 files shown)


charts/capsule/values.yaml
```diff
@@ -4,7 +4,7 @@
 
 manager:
   image:
-    repository: quay.io/clastix/capsule
+    repository: clastix/capsule
     pullPolicy: IfNotPresent
     tag: ''
```
ptx96, line 10: We should also change kubectl [image registry](https://github.com/clastix/capsule/pull/575/files#diff-8facec49e536e52cd2044382bdb9bbd0227328f1977cfee6dc55f73886a67b7bR43), don't we ?

ptx96 (changes requested):
Just a note on kubectl container, otherwise, LGTM!

charts/capsule/values.yaml
prometherion: Nice spot! Created the Docker Hub [repository](https://hub.docker.com/repository/docker/clastix/kubectl), copying in the meanwhile the images from Quay.io to Docker.io!

ptx96 (approved):
LGTM

PR #579, opened 2022-06-02
refactor capsule TLS certificates management

### LINKED ISSUE #578: support cert-manager and capsule controller for generating certificates
in #554 support for `cert-manager` for generating certificates was added. But in this issue, the wrong solution was used, which led to missing `caBundle` in tenant CRD when certificates were generated using `cert-manager`.

We need to develop a single approach for different types of TLS certificates:
- generated by  Capsule controller
- generated by cert-manager
- manually added by user

MaxFedotov:
closes #578 

Added new annotation `capsule.clastix.io/generate-certificates` to `CapsuleConfiguration`. If the value of the annotation is set to `true`, capsule controller will generate and manage ca and TLS certificates.

If it set to `false` capsule controller will only update `caBundle` in webhooks\CRDs with the data present in `capsule-tls` secret.

Merged `ca.crt`, ` tls.crt` and `tls.key` to a single secret in order to make it compatible with the output of `cert-manager`


charts/capsule/Chart.yaml
```diff
@@ -21,7 +21,7 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.1.9
+version: 0.1.10
```
prometherion, line 24: Please, don't bump the chart, we're going to build a new version in a foreseeable short time.

charts/capsule/templates/pre-delete-job.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $cmd := printf "kubectl scale deployment -n $NAMESPACE %s --replicas 0 &&" (include "capsule.deploymentName" .) -}}
 {{- if not .Values.certManager.generateCertificates }}
-{{- $cmd = printf "%s kubectl delete secret -n $NAMESPACE %s %s --ignore-not-found &&" $cmd (include "capsule.secretTlsName" .) (include "capsule.secretCaName" .) -}}
+{{- $cmd = printf "%s kubectl delete secret -n $NAMESPACE %s %s --ignore-not-found &&" $cmd (include "capsule.secretTlsName" .) -}}
```
prometherion, line 3: Remove the second entry means we have to remove the second marker, `%s`.

controllers/tls/manager.go
```diff
@@ -0,0 +1,342 @@
...
+		return true
+	}
+
+	r.Log.Info("Skipping TLS certificate generation as it is still valid")
+
+	return false
+}
+
+// By default helm doesn't allow to use templates in CRD (https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you).
+// In order to overcome this, we are setting conversion strategy in helm chart to None, and then update it with CA and namespace information.
```
prometherion, line 219: Now we got CRD in the `templates` folder, this comment is outdated.

main.go
```diff
@@ -172,116 +142,125 @@ func main() {
 
 	directCfg := configuration.NewCapsuleConfiguration(ctx, directClient, configurationName)
 
-	ca, err := clientset.CoreV1().Secrets(namespace).Get(ctx, directCfg.CASecretName(), metav1.GetOptions{})
-	if err != nil {
-		setupLog.Error(err, "unable to get Capsule CA secret")
+	tlsReconciler := &tlscontroller.Reconciler{
+		Client:        directClient,
```
prometherion, line 146: Using the `directClient` means we're going to query the API server for each read request.

I don't think it's a big problem, here, just want to have feedback from you too.

pkg/webhook/router.go
```diff
@@ -16,12 +15,6 @@ import (
 )
 
 func Register(manager controllerruntime.Manager, webhookList ...Webhook) error {
-	// skipping webhook setup if certificate is missing
-	certData, _ := ioutil.ReadFile("/tmp/k8s-webhook-server/serving-certs/tls.crt")
-	if len(certData) == 0 {
-		return nil
-	}
-
```
prometherion, line 24: If I understood correctly, `controller-manager` will start automatically the webhook server if we got certificates in the right path, isn't it?

controllers/tls/manager.go
MaxFedotov: nope :) as we discussed in slack, moving `crds` to `template` folder will cause this error
```
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: CustomResourceDefinition "tenants.capsule.clastix.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "capsule"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "capsule-system"
helm.go:84: [debug] CustomResourceDefinition "tenants.capsule.clastix.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "capsule"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "capsule-system"
rendered manifests contain a resource that already exists. Unable to continue with update
```
when updating, so this will be backward-incompatible change. That's why we still keep them in `crds` and update `caBundle` for webhook manually

main.go
MaxFedotov: i don't know actually :( but if you think that could be a problem we can use [DelegatingClient](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#NewDelegatingClient) with cacheReader

pkg/webhook/router.go
MaxFedotov: yep, we don't need it because we reconcile certificates before starting the webhook server. And if the reconciliation won't be successful we just do `exit(1)`

MaxFedotov:
> > Merged `ca.crt`, ` tls.crt` and `tls.key` to a single secret in order to make it compatible with the output of `cert-manager`
> 
> Not sure if this is going to be a breaking change for older installations, isn't it?

It would be completely backward-compatible :) during the start of capsule, we reconcile TLS certs before starting controllers/webhook. If the secret with tls certs would miss some of the fields it would just generate a new one and update caBundle in webhooks/crd

main.go
prometherion: No problem at all, let's see in the future if we have to change it, since TLS shouldn't be changes so frequently.

prometherion (changes requested):
LGTM, please just revert the bump up of the Chart since we're going to address it in a future iteration, collecting all the changes made so far.

MaxFedotov:
> LGTM, please just revert the bump up of the Chart since we're going to address it in a future iteration, collecting all the changes made so far.

@prometherion thanks for the review! Requested changes committed

pkg/configuration/client.go
```diff
@@ -62,34 +63,34 @@ func (c capsuleConfiguration) ForceTenantPrefix() bool {
...
-	name = CASecretName
+func (c capsuleConfiguration) TLSSecretName() (name string) {
+	name = TLSSecretName
 
 	if c.retrievalFn().Annotations == nil {
 		return
 	}
 
-	v, ok := c.retrievalFn().Annotations[capsulev1alpha1.CASecretNameAnnotation]
+	v, ok := c.retrievalFn().Annotations[capsulev1alpha1.TLSSecretNameAnnotation]
```
prometherion, line 73: I see that the annotation `CASecretNameAnnotation` is not more used, isn't it?

Formerly, it was used to specify a different Secret name that was containing the CA.

Should we drop it entirely with this refactoring, since we're going to have a single Secret resource containing all the required keys?

prometherion: Just saw the code, and yes, it is no more used: however, we're still mentioning it in several markdown documents for docs.

I'm going to create a new issue for the documentation, since this is still valid with Capsule v0.1.1, drop it from the code-base and everything should be set.

MaxFedotov: @prometherion do i need to remove it from
https://github.com/clastix/capsule/blob/60e826dc832c59b16a3638b161230459681f3a3f/api/v1alpha1/capsuleconfiguration_annotations.go#L8?

MaxFedotov:
@prometherion all done :)

prometherion (approved):

Issue #582, opened 2022-06-07
Grant Global Namespace Patch Rights


oliverbaehler:
# Describe the feature

For users to be able to bootstrap namespaces via Kustomize oder Helm we need global Patch Rights. Because both actions perform a PATCH operation (Dry-Run) on the namespace API when they try to create a namespace. Currently this is not possible:

```
tenants     sample-tenant   False   Namespace/sample-tenant-app-1 dry-run failed, reason: Forbidden, error: namespaces "sample-tenant-app-1" is forbidden: User "system:serviceaccount:tenants:sample-tenant" cannot patch resource "namespaces" in API group "" in the namespace "sample-tenant-app-1"   17h
```

If we implement this functionality we can provide true namespace self service to our developers/customers.
Currently we have a Kyverno Policy which validates this for us, because we grant global PATCH rights to all capsule users. But we have to validate, that patches are only allowed on namespaces which **don't exist** yet or are part of an owned tenant. The policy looks something like this:

```
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: tenant-namespace-edge
spec:
  background: false
  validationFailureAction: "enforce"
  rules:

  - name: disallow-namespace-patch
    context:
    - name: serviceaccount_tenant
      apiCall:
        urlPath: "/api/v1/namespaces/{{serviceAccountNamespace}}/serviceaccounts/{{serviceAccountName}}"
        jmesPath: "not_null(metadata.labels.\"capsule.clastix.io/tenant\")"
    - name: outer_tenant_namespaces
      apiCall:
        urlPath: "/api/v1/namespaces"
        jmesPath: "items[?metadata.labels.\"capsule.clastix.io/tenant\" == \"{{serviceaccount_tenant}}\"].metadata.name"
    match:
      all:
      - resources:
          kinds:
          - Namespace
    exclude:
      any:
      - clusterRoles:
        - cluster-admin
    validate:
      message: "Namespace {{request.object.metadata.name}} has already a date!"
      deny:
        conditions:
          all:

          # Action Must Be Update or Create
          - key: "{{request.operation}}"
            operator: In
            value: ["UPDATE", "CREATE"]

          # If Namespace not in namespaces outer edges (Deny if in) 
          any:  
          - key: "{{request.object.metadata.name}}"
            operator: In
            value: "{{outer_tenant_namespaces}}"
          - key: "{{request.oldObject.metadata.name}}"
            operator: In
[...]

PR #584, opened 2022-06-08
WIP: Grant Global Patch Rights

oliverbaehler:
Implements the following Issue #582 

This is only a Draft to show current Progress.

How can I validate if a namespace is in a tenant, which is owend by the user submitting the API request? Because right now you can still patch any other namespaces in tenants.


oliverbaehler:
I am testing this change with 3 scenarios:

  * **Scenario 1**: Syncs a new namespace and overwrites a existing namespace in the tenant the serviceAccount of the Sync owns.
  * **Scenario 2**: Attempt to Sync a namespace, which does not belong to the tenant the serviceAccount of the Sync owns.
  * **Scenario 3**: Sync a namespace which does not belong to any tenant

**Scenario 1**

Sync works as intended, Flux initially creates a Dry-Run API Request. After that the new namespace is created as expected. The Reconciler runs into a timeout bcs he wants to list the namespaces to verify that they are there (will only work with capsule-proxy)

```
kubectl get kustomization -n tenants
NAME                  READY     STATUS                                                                                                                                                                                                                                                             AGE
sample-tenant-scen1   Unknown   reconciliation in progress                            
```

**Scenario 2**

Fails because the owner hook already detects that the target namespace is not owned by the servieAccount. So we don't need to implement it into the patchWebhook (I am not too sure about that). But the patch is blocked:

```
kubectl get kustomization -n tenants
NAME                  READY     STATUS                                                                                                                                                                                                                                                             AGE                                                                                                                                            14s
sample-tenant-scen2   False     Namespace/sample-tenant2-existing dry-run failed, reason: Cannot assign the desired namespace to a non-owned Tenant, error: admission webhook "owner.namespace.capsule.clastix.io" denied the request: Cannot assign the desired namespace to a non-owned Tenant   14s
```

**Scenario 3**

Admission  Reject when trying to patch `kube-system`

```
kubectl get kustomization -n tenants
NAME                  READY     STATUS                                                                                                                                                                                                                                                             AGE
sample-tenan-scen3    False     Namespace/kube-system dry-run failed, reason: namespace/kube-system can not be patched, error: admission webhook "namespaces.capsule.clastix.io" denied the request: namespace/kube-system can not be patched                                                      14s

```

I have published the scenarios here:

  * https://github.com/oliverbaehler/capsule-flux-tests

prometherion (approved):

PR #598, opened 2022-06-24
feat: update webhooks configuration to inject ca

viveksyngh:
This updates the capsule webhook configurations templates to inject ca-bundle if the certManager is enabled. Otherwise instead of enabling cert manager you would also need to provide custom annotations to enable cert injection.

### DIFF
--- a/charts/capsule/templates/mutatingwebhookconfiguration.yaml
+++ b/charts/capsule/templates/mutatingwebhookconfiguration.yaml
@@ -4,8 +4,11 @@ metadata:
   name: {{ include "capsule.fullname" . }}-mutating-webhook-configuration
   labels:
     {{- include "capsule.labels" . | nindent 4 }}
-  {{- with .Values.customAnnotations }}
   annotations:
+  {{- if .Values.certManager.generateCertificates }}
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "capsule.fullname" . }}-webhook-cert
+  {{-  end }}
+  {{- with .Values.customAnnotations }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
 webhooks:
--- a/charts/capsule/templates/validatingwebhookconfiguration.yaml
+++ b/charts/capsule/templates/validatingwebhookconfiguration.yaml
@@ -4,8 +4,11 @@ metadata:
   name: {{ include "capsule.fullname" . }}-validating-webhook-configuration
   labels:
     {{- include "capsule.labels" . | nindent 4 }}
-  {{- with .Values.customAnnotations }}
   annotations:
+  {{- if .Values.certManager.generateCertificates }}
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "capsule.fullname" . }}-webhook-cert
+  {{-  end }}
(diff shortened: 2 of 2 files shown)


MaxFedotov:
Hi @viveksyngh. When i was implemented #579 i've specially not added cert-manager annotations to webhooks, because capsule is also using webhooks in CRDs and as we can't use helm for templating CRDs capsule will patch webhooks and CRDs manually using certificates generated by cert-manager (https://github.com/clastix/capsule/blob/741db523e5edd8aa0fb5dece5005542217317d9c/controllers/tls/manager.go#L255)

viveksyngh:
@MaxFedotov Then who would add the annotation of the webhook configuration so that cert-manager can populate the ca bundle ? It did not work for me when I deployed in my environment because the annotation on webhooks was missing. Isn't how cert-manager is expected to work or am I missing something?

viveksyngh:
We have seen some race conditions issues with the secret reconciler in our environment which causes webhooks to fail and blocks our deployment or cluster upgrade. We completely want to stop secret reconciler and rely on cert-manager to inject certificates. If required we would separately add annotation on the capsule CRDs. 

If the helm is not managing the CRDs then anyway we need to deploy the CRDs using some other way and we handle the addition of annotation on capsule CRD there.

viveksyngh:
We haven't tested this on the latest version but we have seen this in past. 

I think one of thing we can do is wherever we can not update the `caBundle` with `cert-manager` we can have the `secet-controller` to update that otherwise we should use the `cert-manager`

MaxFedotov:
> We haven't tested this on the latest version but we have seen this in past.

Yes, that's why I've refactored it to try to get rid of this race condition

> I think one of thing we can do is wherever we can not update the caBundle with cert-manager we can have the secet-controller to update that otherwise we should use the cert-manager

Sorry, I can't understand your idea. You want cert-manager to update caBundle in mutating\validating webhooks, while secret-controller will update caBundle in CRD webhook? Am i correct?

viveksyngh:
Yes, that's what I was thinking.

MaxFedotov:
I don't think that it would be a good idea to have 2 different ways to deal with caBundle in different resources. Moreover, we will need to pass additional configuration key to capsule to modify logic of tls-controller. @prometherion WDYT?

prometherion:
I totally agree on that, @MaxFedotov. If you can, @viveksyngh, please give a try the latest changes addressed by Max, on which according to that they should have fixed this race condition.

For further discussions, let's use the [Discussion](https://github.com/clastix/capsule/discussions) section in order to find an agreement that could be even a refactoring of the CA/TLS reconciler.

MaxFedotov:
@prometherion I don't think that this should be merged in a current state. It will lead to two different paths for dealing with certificates

prometherion:
It seems we have to find a consensus here :)

What about discussing this during the next Community Call, scheduled for the 20th of July 2022? We can try to arrange a comfy time for @viveksyngh according to his time zone. Please, ping me on Kubernetes Slack workspace to figure out it!

prometherion:
Mates. may I ask you to get a review of #613? If it looks good to you all, we could close this in favor of the first one.

prometherion:
Closing since #613 has been merged.

PR #601, opened 2022-07-09
Add DaemonSet deploy option in Helm Chart

### LINKED ISSUE #600: Provide daemonset deployment mode in Capsule Helm Chart
# Describe the feature
Deploying Capsule to the cluster is done through the usual means, but for minimal latency in the requests, it might be a good idea to let a `daemonset` deploy the Capsule pods to the same master nodes as the API servers run on.

Since multiple replicas of Capsule is supported, it would be useful to have `daemonset` deploy mode in the Helm Chart too.

The cluster admin uses daemonset deployment mode for Capsule Operator on the same nodes where APIs server is running.

# Expected behavior
Capsule Operator is deployed on the same nodes where APIs server is running

bsctl:
close #600

### DIFF
--- a/charts/capsule/templates/pre-delete-job.yaml
+++ b/charts/capsule/templates/pre-delete-job.yaml
@@ -1,7 +1,4 @@
 {{- $cmd := ""}}
-{{- if eq .Values.manager.kind "Deployment" }}
-{{- $cmd = printf "kubectl scale deployment -n $NAMESPACE %s --replicas 0 &&" (include "capsule.controllerName" .) -}}
-{{- end }}
 {{- if not .Values.certManager.generateCertificates }}
 {{- $cmd = printf "%s kubectl delete secret -n $NAMESPACE %s --ignore-not-found &&" $cmd (include "capsule.secretTlsName" .) -}}
 {{- end }}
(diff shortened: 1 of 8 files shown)


charts/capsule/templates/pre-delete-job.yaml
```diff
@@ -1,4 +1,7 @@
-{{- $cmd := printf "kubectl scale deployment -n $NAMESPACE %s --replicas 0 &&" (include "capsule.deploymentName" .) -}}
+{{- $cmd := ""}}
+{{- if eq .Values.manager.kind "Deployment" }}
+{{- $cmd = printf "kubectl scale deployment -n $NAMESPACE %s --replicas 0 &&" (include "capsule.controllerName" .) -}}
```
bsctl, line 3: @prometherion sorry, not understand the reason to scale down before delete

prometherion: Not sure, I think we can drop it.

prometherion (changes requested):
Let's give a try by removing the scale down upon deletion.

charts/capsule/templates/pre-delete-job.yaml
bsctl: It seems no effects, dropping it.

prometherion (approved):

PR #602, opened 2022-07-09
Grafana dashboard as Config Map

### LINKED ISSUE #599: Create a ready-to-use Grafana dashboard
# Describe the feature
Create a ready-to-use Grafana dashboard in the form of Config Map. Most recent Grafana versions can load automatically dashboards via Config Map. ~~It would be useful to have such CM, including in the Helm Chart, so that no manual dashboard creation is required.~~

bsctl:
close #599


maxgio92:
Just to track the decision. We can provide a `Kustomization` that generate the `ConfigMap` from the `dashboard.json` file, so that the file can still be used "bare" in order to be imported through the Grafana UI.

PR #612, opened 2022-07-21
fix(main.go): generate tls keys before starting the manager when configured

maxgio92:
Signed-off-by: Massimiliano Giovagnoli <me@maxgio.it>


prometherion:
Closing in favor of #613.

PR #613, opened 2022-07-22
Integration with cert-manager

prometherion:
This PR allows deeper integration with `cert-manager` by adding new Helm values.

- `certManager.generateCertificates`: allows to create the Certificate and Issuer requested by `cert-manager`, along with the required annotations for injecting the CA in the Mutating and Validating webhooks
- `tls.create`: allows to skip the creation of the secret, useful if it must be provided externally
- `tls.enableController`: allows to skip the start of the TLS reconciler, useful when `cert-manager` is injecting on its own the CA, or managed by a third party (e.g. Vault, or custom strategies)

Along with that, the `CapsuleConfiguration` has dropped the annotation `capsule.clastix.io/generate-certificates` in favor of `capsule.clastix.io/enable-tls-configuration` that allows controlling the start of the said reconciler. This cannot be considered a breaking change since the annotation hasn't been yet released in a stable version.

### DIFF
--- a/api/v1alpha1/capsuleconfiguration_annotations.go
+++ b/api/v1alpha1/capsuleconfiguration_annotations.go
@@ -8,5 +8,5 @@ const (
 	TLSSecretNameAnnotation                  = "capsule.clastix.io/tls-secret-name"
 	MutatingWebhookConfigurationName         = "capsule.clastix.io/mutating-webhook-configuration-name"
 	ValidatingWebhookConfigurationName       = "capsule.clastix.io/validating-webhook-configuration-name"
-	GenerateCertificatesAnnotationName       = "capsule.clastix.io/generate-certificates"
+	EnableTLSConfigurationAnnotationName     = "capsule.clastix.io/enable-tls-configuration"
 )
--- a/charts/capsule/Makefile
+++ b/charts/capsule/Makefile
@@ -1,0 +1,9 @@
+docs: HELMDOCS_VERSION := v1.8.1
+docs: docker
+	@docker run --rm -v "$$(pwd):/helm-docs" -u $$(id -u) jnorwood/helm-docs:$(HELMDOCS_VERSION)
+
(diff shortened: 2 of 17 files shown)


viveksyngh (approved):
LGTM

MaxFedotov:
@prometherion, I have some questions regarding this PR. 
If I understand correctly, with this PR if I am using cert-manager I had to patch CRDs manually? So there are now two options:
1) Capsule manages certificates - it patches webhooks and CRDs
2) cert-manager manages certificates - it patches only webhooks and i had to patch CRDs manually?

prometherion:
@MaxFedotov if you're enabling Cert Manager, CRD must be patched manually.

However, you can still run the TLS reconcile with the Helm value `tls.enableController` that will inject the CA in CRD and webhooks, although creating a sort of conflict that is not breaking.

MaxFedotov:
@prometherion What do you think if we skip reconciling certificates if `certManager.generateCertificates` and `tls.enableController` are both enabled? and will only patch CRDs\Webhooks in order to prevent all possible conflicts (e.g. what will happen when a certificate will expire and both, Capsule and cert-manager will try to update it? better to prevent this condition at all)?

prometherion:
@MaxFedotov I think it's up to the cluster administrator, we're providing all the available options to play the desired implementation:

1. creating Secrets and generating certificates by Capsule
2. let cert-manager generate certificates, by creating or non creating Secret resources, and enabling or disabling the TLS reconciler
3. externally managed Secret for CA and TLS, without TLS reconciler

All of these options can be easily put in place with the new Helm values, translated in `CapsuleConfiguration` annotations, and Helm resource creation.

Since these are advanced parameters, they're expected to use just by people who are aware of what they're doing.

MaxFedotov:
lgtm!

MaxFedotov (approved):
LGTM

PR #629, opened 2022-08-03
Added makefile command to autogen CRDs documentation

mastrogiovanni:
I added a Makefile command to create a reference.md file into docs starting from CRDs from config/crd/base.
The tool chosen is a result of several trials since:
- standard kubernetes tools requires connection to API Server
- stardard kubernetes reference libraries works with openAPIv2 but not v3 (that is incomplete)
- other tools (in python) ones seems incomplete and buggy and can introduce some problem in maintenance of the codebase
- the tool chosen seems to belong to a cloud native platform with a good track record


config/crd/bases/capsule.clastix.io_capsuleconfigurations.yaml
```diff
@@ -1,4 +1,3 @@
-
```
prometherion, line 1: Why this first line has been removed? Wondering about this because these files are auto-generated.

Makefile
```diff
@@ -147,6 +150,10 @@ CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
 	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0)
 
+APIDOCS_GEN = $(shell pwd)/bin/crdoc
+apidocs-gen: ## Download controller-gen locally if necessary.
```
prometherion, line 154: Typo on `Download controller-gen locally if necessary`

Makefile
```diff
@@ -78,6 +78,9 @@ manifests: controller-gen
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+apidoc: apidocs-gen
+	$(APIDOCS_GEN) crdoc --resources config/crd/bases --output docs/content/general/references.md
```
prometherion, line 82: Rather than pointing to `docs/content/general/references.md`, couldn't we use a template to reference the resulting file in the `references.md` file that is useful since providing further information about Capsule?

e.g.:

```go
# Reference

Reference document for Capsule Operator configuration

## Custom Resource Definition

{{ .tenantCRDMarkdown }}

## Capsule Configuration

The Capsule configuration can be piloted by a Custom Resource definition named `CapsuleConfiguration`.

{{ .capsuleConfigurationCRDMarkdown }}
````

**Review by prometherion — CHANGES_REQUESTED** [review]
The generated files must be committed, each commit must be atomic and ensuring all the auto-generated files are up to date.

File: config/crd/bases/capsule.clastix.io_capsuleconfigurations.yaml (hunk 1 above)
Comment (mastrogiovanni) [reply config/crd/bases/capsule.clastix.io_capsuleconfigurations.yaml:1]: My bad

File: config/crd/bases/capsule.clastix.io_tenants.yaml [hunk 4]
```diff
@@ -1,4 +1,3 @@
-
```
prometherion, line 1: Same here, please, keep the empty new line at the beginning if this has been generated by `controller-gen`

mastrogiovanni:
I reset the `references.md` file to the old version and I just added the newly generated `tenantCRDMarkdown.md` to the PR. Not clear to me if `references.md` need to be autogenerated or not. I think that having a section for CRD and another for the rest is the best strategy.

prometherion:
> I think that having a section for CRD and another for the rest is the best strategy.

Absolutely, let's remove the CRD scheme from `docs/content/general/references.md` file, pointing with a link to the new file.

Remember to add it to the side menu, suddenly after `References` with the title `APIs`.

prometherion:
The output rendered is not so superb: https://deploy-preview-629--capsule-documentation.netlify.app/docs/general/tenant-crd#capsuleconfiguration.

Could we try to customize it a bit, by removing the `↩ Parent` link?

mastrogiovanni:
I customized the template that originate the page and added some styling to improve table readibility

prometherion (approved):

PR #631, opened 2022-08-13
WIP: feat/gitops ready rbac

### LINKED ISSUE #608: Improve GitOps user experience
> ### This is a draft of the proposal.

### Describe the feature

It would be useful to let Capsule set all the needed RBAC and identities for a machine Tenant Owner, that would be responsible of reconciling Tenant resources. 

### What would the new user story look like?

As a platform/admin user, I want to provide tenant users the ability to declare their resources and let a GitOps Kubernetes operator reconcile them from Git, with an unprivileged identity.

#### RBAC

By using a knob (e.g. `spec.gitopsReady=true`) the admin user could request that the Tenant would be prepared to have:
- Tenant Owner as `ServiceAccount`, to be used by GitOps operators and placed into a dedicated Namespace ("Tenant Owner Home")
- necessary (`Cluster`)`Role`s bound, in order to patch, work with custom resources inside the Tenant (admin ClusterRole is not enough, cluster-admin is needed)
- (optional) `ClusterRole` bound in order to impersonate himself (e.g. needed on Flux use case
- (optional) `Secret` with generated kubeconfig to let the Tenant Owner SA communicate over Capsule Proxy 

I'm going to refine this propsal with more detail and the rationale behind it.

#### Client config for Proxy

For each robot Tenant Owner (i.e. `ServiceAccount` type) a `Secret` containing `kubeconfig` to configure a client to communicate as Tenant Owner through the Capsule Proxy.

#### Related issues
- #528

maxgio92:
Closes #608

### DIFF
--- a/api/v1beta1/owner_role.go
+++ b/api/v1beta1/owner_role.go
@@ -38,7 +38,20 @@ func (in OwnerSpec) GetRoles(tenant Tenant, index int) []string {
 		}
 	}
 
-	return []string{"admin", "capsule-namespace-deleter"}
+	roles := []string{"admin", "capsule-namespace-deleter"}
+
+	if tenant.Spec.GitOpsReady {
+		roles = append(roles, in.getGitOpsRoles(tenant)...)
+	}
+
+	return roles
+}
+
+func (in OwnerSpec) getGitOpsRoles(tenant Tenant) []string {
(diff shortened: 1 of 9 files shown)


maxgio92:
Won't do. Please read #608

PR #645, opened 2022-09-26
docs: clarify serviceAccounts as tenant owner

lalyos:
after some discussion on slack with @prometherion here is my attempt to make the docs more clear.

> we don't know how's the platform admin is managing these ServiceAccount
We saw from other end-users that their best practice is to create a Namespace containing all the ServiceAccount resources that will be used as Tenant Owner (e.g: tenant-system) in order to user its group in the Capsule Configuration (e.g.: system:serviceaccounts:tenant-system).

### DIFF
--- a/docs/content/general/tutorial.md
+++ b/docs/content/general/tutorial.md
@@ -139,15 +139,15 @@ metadata:
   name: oil
 spec:
   owners:
-  - name: system:serviceaccount:default:robot
+  - name: system:serviceaccount:tenant-system:robot
     kind: ServiceAccount
 EOF
 ```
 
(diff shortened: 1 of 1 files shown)


prometherion (changes requested):
LGTM, just a note regarding the commit message that is violating the semantic commit messages.

I would suggest something has `fix(docs): clarifying serviceaccount as tenant owner`

lalyos:
fixed the commit message

bsctl (approved):
@lalyos thanks for your work, it lgtm.

prometherion (approved):

Issue #633, opened 2022-08-16
Support for RuntimeClass


prometherion:
# Describe the feature

[RuntimeClass](https://kubernetes.io/docs/concepts/containers/runtime-class/) is a feature for selecting the container runtime configuration. The container runtime configuration is used to run a Pod's containers.

# What would the new user story look like?

As a Cluster Administrator, I want to limit, prevent, or allow the use of specific `RuntimeClass` to specific tenants.

# Expected behaviour

When the Tenant is enforced with a specific set of `RuntimeClass`, if the Pod is using a forbidden one, must be denied.

oliverbaehler:
@prometherion I ll do that. I guess we are looking for same functionality as eg. PriorityClasses:

```
  runtTimeClass:
    default: "cri-o"
    allowed:
      - default
    allowedRegex: "^docker-.*$"
```

I would also add the `default` attribute and could combine it with #610. That  makes sense for you?

prometherion:
> I guess we are looking for same functionality as eg. PriorityClasses:

I confirm that, also because we do already have our mechanism to check regex values and the exact match.

> I would also add the `default` attribute and could combine it with #610. That makes sense for you?

Let's keep it separated to avoid huge PRs.

Just a minor note, we have to address this for the new type, `v1beta2`, and we're still blocked since we don't have the required scaffolding.

I'll be back to you once everything is settled so you can start working on it! 💪🏻

prometherion:
@oliverbaehler I missed to assign this to you, wondering if you're still on it, otherwise I can take care.

prometherion:
Taking over of this as discussed privately.

PR #648, opened 2022-09-29
feat: add runtimeclass control

oliverbaehler:
Signed-off-by: Oliver Baehler <oliver.baehler@hotmail.com>

This PR resolves multiple bugs/features:
- Implements #633 
- Corrects local Development commands in Makefile
- Validate on Image Pod Updates Registry according to tenant configuration

Currently under development, have to test everything tomorrow.

### DIFF
--- a/docs/content/general/tutorial.md
+++ b/docs/content/general/tutorial.md
@@ -838,6 +838,42 @@ With the said Tenant specification, Alice can create a Pod resource if `spec.pri
 
 If a Pod is going to use a non-allowed _Priority Class_, it will be rejected by the Validation Webhook enforcing it.
 
+
+## Assign Pod Runtime Classes
+
+Pods can be assigned different runtime classes. With the assigned runtime you can control Container Runtime Interface (CRI) is used for each pod. See [Kubernetes documentation](https://kubernetes.io/docs/concepts/containers/runtime-class/). 
+
+To prevent misuses of Pod Runtime Classes, Bill, the cluster admin, can enforce the allowed Pod Runtime Class at tenant level:
+
+```yaml
+kubectl apply -f - << EOF
+apiVersion: capsule.clastix.io/v1beta2
+kind: Tenant
+metadata:
+  name: oil
+spec:
+  owners:
+  - name: alice
+    kind: User
+  RuntimeClasses:
+    allowed:
+    - legacy
+    allowedRegex: "^hardened-.*$"
+    selector:
+      matchLabels:
+        env: "production"
+EOF
+```
+
+With the said Tenant specification, Alice can create a Pod resource if `spec.RuntimeClasses` equals to:
+
+- `legacy`
+- `hardened-crio` or `hardened-containerd`, since these compile the allowed regex.
+- Any RuntimeClass which has the label `env` with the value `production`
+
+If a Pod is going to use a non-allowed _Runtime Class_, it will be rejected by the Validation Webhook enforcing it.
+
+
 ## Assign Nodes Pool
 Bill, the cluster admin, can dedicate a pool of worker nodes to the `oil` tenant, to isolate the tenant applications from other noisy neighbors.
 
--- a/docs/content/general/tutorial.md
+++ b/docs/content/general/tutorial.md
@@ -855,7 +855,7 @@ spec:
   owners:
   - name: alice
     kind: User
-  RuntimeClasses:
+  runtimeClasses:
     allowed:
     - legacy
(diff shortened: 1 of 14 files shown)


prometherion (commented):
This feature is definitely useful and must be landed!

However, adding a new field to an already published CRD version is not the right way, since it breaks the contract with the previous installation.

I'd ask you to wait for #644 for getting merged and land the `runtimeClass` to the `v1beta2` API group, rather than `v1alpha1`.

And at the same time, maybe, you could decouple the commits regarding optimising the dev environment since these have been already addressed with #650.

prometherion:
@oliverbaehler may I ask you to rebase over the branch `release/v0.2.0` and add this feature to the `v1beta2` API type, please?

pkg/webhook/pod/runtimeclass.go
```diff
@@ -0,0 +1,114 @@
...
+	if len(name) == 0 {
+		return nil, nil
+	}
+
+	obj := &nodev1.RuntimeClass{}
+
+	if err := c.Get(ctx, types.NamespacedName{Name: name}, obj); err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
+		}
```
prometherion, line 40: TIL that a Pod cannot be created using a non-existing Runtime Class: we can drop the `IsNotFound` error check and return it.

pkg/webhook/pod/runtimeclass.go
```diff
@@ -0,0 +1,114 @@
...
+func (h *runtimeClass) OnDelete(client.Client, *admission.Decoder, record.EventRecorder) capsulewebhook.Func {
+	return func(ctx context.Context, req admission.Request) *admission.Response {
+		return nil
+	}
+}
+
+func (h *runtimeClass) OnUpdate(client.Client, *admission.Decoder, record.EventRecorder) capsulewebhook.Func {
+	return func(ctx context.Context, req admission.Request) *admission.Response {
+		return nil
+	}
```
prometherion, line 63: Can be a pod updated with a newer runtime class?

prometherion (changes requested):

prometherion (changes requested):
Please, copy the `CustomResourceDefinition` for the Tenant resource into the `charts/capsule/crds` folder so we keep it updated and coherent with the changes introduced by this PR.

pkg/webhook/pod/runtimeclass.go
oliverbaehler: These are the only field we need to worry about regarding pod updates:

```
spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds`, `spec.tolerations` (only additions to existing tolerat>
```

pkg/webhook/pod/runtimeclass.go
oliverbaehler: You are right. will remove:

```
kubectl apply -f sample.yaml 
Error from server (Forbidden): error when creating "sample.yaml": pods "nginx" is forbidden: pod rejected: RuntimeClass "myclass" not found
```

pkg/webhook/pod/runtimeclass.go
prometherion: That's great, this simplifies the handling of the webhook, thanks!

docs/content/general/tutorial.md
```diff
@@ -838,6 +838,42 @@ With the said Tenant specification, Alice can create a Pod resource if `spec.pri
...
+kubectl apply -f - << EOF
+apiVersion: capsule.clastix.io/v1beta2
+kind: Tenant
+metadata:
+  name: oil
+spec:
+  owners:
+  - name: alice
+    kind: User
+  RuntimeClasses:
```
prometherion, line 858: It must be `runtimeClasses` as the JSON struct tag

pkg/webhook/pod/runtimeclass.go
```diff
@@ -0,0 +1,108 @@
...
+
+		return &response
+	}
+
+	switch {
+	case allowed == nil:
+		// Enforcement is not in place, skipping it at all
+		return nil
+	case len(runtimeClassName) == 0:
+		// We don't have to force Pod to specify a Priority Class
```
prometherion, line 97:
```suggestion
		// We don't have to force Pod to specify a RuntimeClass
```

prometherion (changes requested):

prometherion (approved):

PR #651, opened 2022-10-05
Enforcing container registry also upon Pod update

oliverbaehler:
Since Images can by patched on pod manifests, we must also validate patch operations on pods regarding the container registry. If not, we can bypass the container registry, since the use is able to update the image on container level. So I will add that the container registry is validated on update

This fixes a serious security issues and should be released asap.


e2e/container_registry_test.go
```diff
@@ -110,6 +110,34 @@ var _ = Describe("enforcing a Container Registry", func() {
...
+				},
+			},
+		}
+
+		cs := ownerClient(tnt.Spec.Owners[0])
+		_, err := cs.CoreV1().Pods(ns.Name).Create(context.Background(), pod, metav1.CreateOptions{})
+		if err == nil {
+			c := pod.DeepCopy()
+			c.Spec.Containers[0].Image = "attacker/google-containers/pause-amd64:3.0"
+			_, err := cs.CoreV1().Pods(ns.Name).Update(context.Background(), c, metav1.UpdateOptions{})
```
oliverbaehler, line 136: @prometherion I need your help here, i don't get it running. I just want to simulate, that a capsule user patches the image of a pod with an attacker registry. But it doesn't work.

prometherion: I'll try to debug your upstream ASAP and get back to you! 👍🏻

prometherion: May I ask you to rebase over the latest changes on `master` and ensure first the manifests are aligned?

prometherion:
Implemented with #648.

PR #673, opened 2022-12-26
First implementation for label selector for Ingress, Storage and Priority classes

### LINKED ISSUE #436: Use selectors instead of AllowedList for IngressClasses, StorageClasses, PriorityClasses
We are using a specific struct, `AllowedListSpec`:
```go
type AllowedListSpec struct {
	Exact []string `json:"allowed,omitempty"`
	Regex string   `json:"allowedRegex,omitempty"`
}
```
to allow tenant users to use only specific IngressClasses, StorageClasses or PriorityClasses. 

While it allows additional fine-grained configuration, that is not a kubernetes way of specifying resources. 
Much better will be to use label selectors (the same way, as it is in `nodeSelector` option for tenant).

With this, `tenant spec` will look like:
```yaml
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - name: alice
    kind: User
  storageClasses:
    selector:
       foo: bar  
```

This change will simplify webhook code, make it works much faster (no regexp check), and will simplify the code of a `capsule-proxy` (as we won't need to get all resources first and then perform regexp validation or array probing).

The downside of this change is that it is completely backward incompatible and there will be no way to automatically upgrade tenant spec to this new version.

@bsctl, @prometherion need your opinion on this.

prometherion:
Closes #436.

This feature cannot be backported to `v1beta1` and all the Tenant instances must be persisted first as `v1beta2` due to the complexity of storing the `matchLabels` and the `matchRequirements` data structure.

Aong with that, `v1beta2` is marked as the storage version for the upcoming Capsule release.


Makefile
```diff
@@ -172,6 +179,14 @@ GINKGO = $(shell pwd)/bin/ginkgo
 ginkgo: ## Download ginkgo locally if necessary.
 	$(call go-install-tool,$(GINKGO),github.com/onsi/ginkgo/ginkgo@v1.16.5)
 
+CT = $(shell pwd)/bin/ct
+ct: ## Download ct locally if necessary.
+	$(call go-install-tool,$(CT),github.com/helm/chart-testing/v3/ct@v3.7.1)
+
+KIND = $(shell pwd)/bin/kind
+kind: ## Download kind locally if necessary.
+	$(call go-install-tool,$(CT),sigs.k8s.io/kind/cmd/kind@v0.17.0)
```
prometherion, line 188: Shouldn't be `$(KIND)` instead of `$(CT)`?

Makefile
```diff
@@ -86,8 +86,15 @@ helm-docs: HELMDOCS_VERSION := v1.11.0
 helm-docs: docker
 	@docker run -v "$(SRC_ROOT):/helm-docs" jnorwood/helm-docs:$(HELMDOCS_VERSION) --chart-search-root /helm-docs
 
-helm-lint: docker
-	@docker run -v "$(SRC_ROOT):/workdir" --entrypoint /bin/sh quay.io/helmpack/chart-testing:v3.3.1 -c "cd /workdir && ct lint --config .github/configs/ct.yaml --lint-conf .github/configs/lintconf.yaml --all --debug"
+helm-lint: ct
+	@ct lint --config $(SRC_ROOT)/.github/configs/ct.yaml --lint-conf $(SRC_ROOT)/.github/configs/lintconf.yaml --all --debug
+
+helm-install: kind ct docker-build
```
prometherion, line 92: Maybe instead of `helm-install` we should use `helm-test`? At first sight, with that name I was expecting having helm installed locally.

oliverbaehler (approved):
LGTM

Issue #674, opened 2022-12-28
Enforcing container registry also upon Pod update


prometherion:
Since Images can be patched on pod manifests, we must also validate patch operations on pods regarding the container registry. If not, we can bypass the container registry, since the user is able to update the image on the container level.

prometherion:
Closed via #648.

prometherion:
Closes via #675.

PR #675, opened 2022-12-29
fix: reject registry patches & v1beta2 fixes

oliverbaehler:
* Implements #674 
* Fixes reconciler must use v1beta2
* Fixes local development (auto patching of conversion webhook paths)


oliverbaehler:
@prometherion I was thinking if we should also add a option on tenant basis to prevent image patches in a tenant

Story:
As a tenant owner I want to disallow patches of images for pods.

prometherion:
> @prometherion I was thinking if we should also add a option on tenant basis to prevent image patches in a tenant. What do you think?
> 
> Story: As a tenant owner I want to disallow patches of images for pods.

I'm trying to understand why an update should be allowed or not, also considering that now, with the fix, the pod container registry is enforced at every level.

tl;dr; what's the reason in allowing the patch of images?

oliverbaehler:
@prometherion  Good question, I guess sometimes for development reasons or something like. I used it once or twice when solving production issues with a static pod to copy some files etc.

The question is, if we really want to deny it per default. That would be a change to the default behavior for end users and I don't know if thats what we are looking for as a framework.

prometherion:
I think there's nothing wrong until you use an image using the same container registry, thus, trusted.

Honestly, I don't see any need for a such feature, we can wait for feedback requests with a more detailed user story driven by business requirements.

prometherion (approved):
Astonishing work, Oliver, this kicks ass! 🚀

Issue #610, opened 2022-07-21
Default Value PriorityClass/storageClasses/allowedClasses


oliverbaehler:
# Describe the feature

Generally it's a default value for `priorityClasses`, `storageClasses` and `allowedClasses` (Ingress). Adding a default value would simplify working in the tenant for the customer, since they wouldn't have to worry to much about these values, since a default value is applied. But then con overwrite it, by setting their value.

Would you be happy to accept this feature(s)? I will do the contribution.

# What would the new user story look like?

Story for PriorityClasses.

Add the feature to define a default PriorityClass for all pods within the tenant. The given default PriorityClass would be assigned to all pods, which don't have a priorityClass set. If a PriorityClass is set on a pad, the default is ignored. The default value is also verified against the allowed/regex.

A new tenant is created. The customer can assign two tiers `tier-gold` and `tier-silver`. To make it more convenient to user only has to tag pods with higher priority with `tier-gold`. While all the other pods are assigned the `tier-silver`. Manifest could look like this:

```
apiVersion: capsule.clastix.io/v1beta1
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - name: alice
    kind: User
  priorityClasses:
    default: "tier-silver"
    allowed:
    - default
    allowedRegex: "^tier-.*$"
```

oliverbaehler:
Maybe it would even make sense adding a namespace selector.


```
apiVersion: capsule.clastix.io/v1beta1
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - name: alice
    kind: User
  priorityClasses:
    defaults:
      - value: "tier-prod"
         namespaceSelector:
            matchExpressions:
              - {key: environment, operator: NotIn, values: [dev]}
      - value: "tier-silver"
         namespaceSelector:
            matchExpressions:
              - {key: environment, operator: In, values: [dev]}
    allowed:
    - default
    allowedRegex: "^tier-.*$"
```

prometherion:
This feature would require a mutating webhook in order to patch pods with the default value.

Along with that, I would start using a bare value to keep the homogeneity of the other keys such as `allowed` that are referring to bare names.

Regarding the selector, it's something we are already evaluating in #436, and that would be absolutely breaking since it means dropping the exact match and the regex one.

prometherion:
We can work on this, once #644 is merged.

oliverbaehler:
@prometherion you can assign it to me, i have already some code for this feature but will wait for the new apiv

[...]

PR #678, opened 2023-01-05
feat: defaults handler for ingress, pods and pvcs

oliverbaehler:
This Pull Request implements #610 

Required some code refactoring to keep code as clean as possible.

Here's some notes for the review:

- The assumption is always, that if a value for `Default` is defined, it's automatically allowed within a tenant. The default value does not have to be explicit allowed (Better User-Experience).
- Patches are only mutated for `Ingress`. `PVCs` and `Pods` can not be patched on the relevant fields, which would cause a default mutation.
- `IngressClass` only supported on API `networking/v1`. Feature supported for `Ingress` on `networking` `v1` and `v1beta1`
- `StorageClass` only supported on API `storage/v1`
- `PriorityClass` only supported on API `schedule/v1`


prometherion:
> * `PVCs` and `Pods` can not be patched on the relevant fields, which would cause a default mutation.

I'm not getting the point here: if we define a mutating webhook for these fields, we should be able to respond with a mutating response that would be executed by the API Server, as we're doing with Namespaces.

Am I missing something else?

pkg/webhook/route/defaults.go
```diff
@@ -0,0 +1,28 @@
...
+
+package route
+
+import (
+	capsulewebhook "github.com/clastix/capsule/pkg/webhook"
+)
+
+// +kubebuilder:webhook:path=/defaults,mutating=true,sideEffects=None,admissionReviewVersions=v1,failurePolicy=fail,groups="",resources=pods,verbs=create,versions=v1,name=pod.defaults.capsule.clastix.io
+// +kubebuilder:webhook:path=/defaults,mutating=true,sideEffects=None,admissionReviewVersions=v1,failurePolicy=fail,groups="",resources=persistentvolumeclaims,verbs=create,versions=v1,name=storage.defaults.capsule.clastix.io
+// +kubebuilder:webhook:path=/defaults,mutating=true,sideEffects=None,admissionReviewVersions=v1,failurePolicy=fail,groups=networking.k8s.io,resources=ingresses,verbs=create;update,versions=v1beta1;v1,name=ingress.defaults.capsule.clastix.io
```
prometherion, line 12: We can use a single annotation and separate groups with the `;` symbol: this would lead to a single new webhook in the `mutatingwebhookconfiguration` resource and reduce the possible human error when aligning the Kustomize and the Helm chart.

oliverbaehler: But how do you want to deal with the fact, that only ingresses use the `update` verb, but the others don't?

prometherion: Thanks for pointing that, you're right. Then, let's keep 3 different rules for that.

api/v1beta2/tenant_types.go
```diff
@@ -18,7 +18,7 @@ type TenantSpec struct {
 	// Specifies options for the Service, such as additional metadata or block of certain type of Services. Optional.
 	ServiceOptions *api.ServiceOptions `json:"serviceOptions,omitempty"`
 	// Specifies the allowed StorageClasses assigned to the Tenant. Capsule assures that all PersistentVolumeClaim resources created in the Tenant can use only one of the allowed StorageClasses. Optional.
```
prometherion, line 20: Should we maybe also update the documentation, stating:
```
Specifies the allowed StorageClasses assigned to the Tenant. Capsule assures that all PersistentVolumeClaim resources created in the Tenant can use only one of the allowed StorageClasses. A default value can be specified, and all the PersistentVolumeClaim resources created will inherit the declared class. Optional.
```

api/v1beta2/tenant_types.go
```diff
@@ -38,7 +38,7 @@ type TenantSpec struct {
 	// Specifies the allowed RuntimeClasses assigned to the Tenant. Capsule assures that all Pods resources created in the Tenant can use only one of the allowed RuntimeClasses. Optional.
 	RuntimeClasses *api.SelectorAllowedListSpec `json:"runtimeClasses,omitempty"`
 	// Specifies the allowed priorityClasses assigned to the Tenant. Capsule assures that all Pods resources created in the Tenant can use only one of the allowed PriorityClasses. Optional.
```
prometherion, line 40: Same discussion with the default value.

api/v1beta2/tenant_types.go
```diff
@@ -18,7 +18,7 @@ type TenantSpec struct {
 	// Specifies options for the Service, such as additional metadata or block of certain type of Services. Optional.
 	ServiceOptions *api.ServiceOptions `json:"serviceOptions,omitempty"`
 	// Specifies the allowed StorageClasses assigned to the Tenant. Capsule assures that all PersistentVolumeClaim resources created in the Tenant can use only one of the allowed StorageClasses. Optional.
-	StorageClasses *api.SelectorAllowedListSpec `json:"storageClasses,omitempty"`
+	StorageClasses *api.DefaultAllowedListSpec `json:"storageClasses,omitempty"`
 	// Specifies options for the Ingress resources, such as allowed hostnames and IngressClass. Optional.
```
prometherion, line 22: Same discussion with the default value.

charts/capsule/templates/validatingwebhookconfiguration.yaml
```diff
@@ -259,32 +287,4 @@ webhooks:
...
-        - ""
-      apiVersions:
-        - v1
-      operations:
-        - UPDATE
-      resources:
-        - nodes
-  sideEffects: None
-  timeoutSeconds: {{ .Values.validatingWebhooksTimeoutSeconds }}
+  timeoutSeconds: {{ .Values.validatingWebhooksTimeoutSeconds }}
```
prometherion, line 290: Missing new line at the end of the file.

config/webhook/patch_mutating_ns_selector.yaml
```diff
@@ -0,0 +1,27 @@
...
+        operator: Exists
+- op: add
+  path: /webhooks/0/rules/0/scope
+  value: Namespaced
+- op: add
+  path: /webhooks/1/rules/0/scope
+  value: Namespaced
+- op: add
+  path: /webhooks/2/rules/0/scope
+  value: Namespaced
```
prometherion, line 27: Missing new line at the end of the file.

e2e/pod_priority_class_test.go
```diff
@@ -209,4 +299,162 @@ var _ = Describe("enforcing a Priority Class", func() {
 		}
 	})
 
+	// Ensure default field does not cause false positives for the previous tests
+	It("add default to tenant", func() {
+		tnt.Spec.PriorityClasses.Default = "tenant-default"
+	})
```
prometherion, line 305: This test assertion is useless, since there's no update at all: please, fix it with an update in the `It` block or drop it.

e2e/storage_class_test.go
```diff
@@ -51,9 +98,18 @@ var _ = Describe("when Tenant handles Storage classes", func() {
 	})
 	JustAfterEach(func() {
 		Expect(k8sClient.Delete(context.TODO(), tnt)).Should(Succeed())
+		Eventually(func() (err error) {
+			err = k8sClient.DeleteAllOf(context.TODO(), &storagev1.StorageClass{}, &client.DeleteAllOfOptions{})
+			if err != nil && !k8serrors.IsNotFound(err) {
+				return err
+			}
+			return nil
```
prometherion, line 106: Let's delete all the StorageClass resources marked with the label selector `has env`, please, keeping the default one created by `kind`.

e2e/storage_class_test.go
```diff
@@ -149,5 +244,185 @@ var _ = Describe("when Tenant handles Storage classes", func() {
...
+					return err
+				}).Should(Succeed())
+			}
+		})
+	})
+
+	// Ensure default field does not cause false positives for the previous tests
+	It("add default to tenant", func() {
+		tnt.Spec.StorageClasses.Default = "tenant-default"
+	})
```
prometherion, line 290: Same here: define automatically the default value in the `tnt` variable assignment.

pkg/api/allowed_list.go
```diff
@@ -15,18 +15,49 @@ import (
...
+type DefaultAllowedListSpec struct {
+	SelectorAllowedListSpec `json:",inline"`
+	Default                 string `json:"default,omitempty"`
+}
+
+func (in *DefaultAllowedListSpec) MatchDefault(value string) bool {
+	if in.Default != "" {
+		if in.Default == value {
+			return true
+		}
```
prometherion, line 27:
```suggestion
		return in.Default == value
```

pkg/api/allowed_list.go
```diff
@@ -15,18 +15,49 @@ import (
...
 type SelectorAllowedListSpec struct {
 	AllowedListSpec      `json:",inline"`
 	metav1.LabelSelector `json:",inline"`
 }
 
+func (in *SelectorAllowedListSpec) MatchSelectByName(obj client.Object) bool {
+	if obj != nil {
+		if in.AllowedListSpec.Match(obj.GetName()) || in.SelectorMatch(obj) {
+			return true
+		}
```
prometherion, line 44:
```suggestion
		return in.AllowedListSpec.Match(obj.GetName()) || in.SelectorMatch(obj)
```

pkg/webhook/defaults/pods.go
```diff
@@ -0,0 +1,106 @@
...
+		if priorityClassPod != allowed.Default {
+			cpc, err = utils.GetPriorityClassByName(ctx, c, priorityClassPod)
+			// Should not happen, since API already checks if PC present
+			if err != nil {
+				response := admission.Denied(NewPriorityClassError(priorityClassPod, err).Error())
+
+				return &response
+			}
+		}
+	} else {
```
prometherion, line 65:
```suggestion
	if priorityClassPod != "" && priorityClassPod != allowed.Default {
		cpc, err = utils.GetPriorityClassByName(ctx, c, priorityClassPod)
		// Should not happen, since API already checks if PC present
		if err != nil {
			response := admission.Denied(NewPriorityClassError(priorityClassPod, err).Error())

			return &response
		}
	} else {
```

pkg/webhook/defaults/pods.go
```diff
@@ -0,0 +1,106 @@
...
+		mutate = true
+	}
+
+	// If given class is cluster default
+	if cpc != nil && utils.IsDefaultPriorityClass(cpc) {
+		// If cluster default is not allowed in tenant
+		if !tnt.Spec.PriorityClasses.MatchSelectByName(cpc) {
+			mutate = true
+		}
+	}
```
prometherion, line 75:
```suggestion
	// If given class is cluster default
	if cpc != nil && utils.IsDefaultPriorityClass(cpc) && !tnt.Spec.PriorityClasses.MatchSelectByName(cpc) {
		// If cluster default is not allowed in tenant
		mutate = true
	}
```

pkg/webhook/defaults/pods.go
```diff
@@ -0,0 +1,106 @@
...
+			response := admission.Denied(msg)
+
+			return &response
+		}
+
+		pod.Spec.PreemptionPolicy = pc.PreemptionPolicy
+		pod.Spec.Priority = &pc.Value
+		pod.Spec.PriorityClassName = pc.Name
+
+		recorder.Eventf(tnt, corev1.EventTypeNormal, "TenantDefault", "Assigned Tenant default Priority Class %s to %s/%s", allowed.Default, pod.Namespace, pod.Name)
```
prometherion, line 92: Shouldn't we record this once we're sure the handler exited without any error?

At line 97 there's still a chance of getting an error.

pkg/webhook/defaults/ingress.go
```diff
@@ -0,0 +1,89 @@
...
+	if ingressClassName != nil {
+		if ingressClassName != &allowed.Default {
+			ingressClass, err = utils.GetIngressClassByName(ctx, c, ingressClassName)
+			if err != nil && !k8serrors.IsNotFound(err) {
+				response := admission.Denied(NewIngressClassError(*ingressClassName, err).Error())
+
+				return &response
+			}
+		}
+	} else {
```
prometherion, line 61:
```suggestion
	if ingressClassName != nil && ingressClassName != &allowed.Default {
		ingressClass, err = utils.GetIngressClassByName(ctx, c, ingressClassName)
		if err != nil && !k8serrors.IsNotFound(err) {
			response := admission.Denied(NewIngressClassError(*ingressClassName, err).Error())
	
			return &response
		}
	} else {
```

pkg/webhook/defaults/ingress.go
```diff
@@ -0,0 +1,89 @@
...
+		mutate = true
+	}
+
+	// If given class is cluster default
+	if ingressClass != nil && utils.IsDefaultIngressClass(ingressClass) {
+		// If cluster default is not allowed in tenant
+		if !allowed.MatchSelectByName(ingressClass) {
+			mutate = true
+		}
+	}
```
prometherion, line 71:
```suggestion
	// If given class is cluster default
	if ingressClass != nil && utils.IsDefaultIngressClass(ingressClass) && !allowed.MatchSelectByName(ingressClass) {
		// If cluster default is not allowed in tenant
		mutate = true
	}
```

pkg/webhook/defaults/ingress.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if mutate {
+		ingress.SetIngressClass(allowed.Default)
+		recorder.Eventf(tnt, corev1.EventTypeNormal, "TenantDefault", "Assigned Tenant default Ingress Class %s to %s/%s", allowed.Default, ingress.Name(), ingress.Namespace())
```
prometherion, line 75: We should record only if the handler is returning with no error, still a chance for an error at line 80.

pkg/webhook/defaults/storage.go
```diff
@@ -0,0 +1,90 @@
...
+	if storageClassName != nil {
+		if storageClassName != &allowed.Default {
+			csc, err = utils.GetStorageClassByName(ctx, c, *storageClassName)
+			if err != nil && !k8serrors.IsNotFound(err) {
+				response := admission.Denied(NewStorageClassError(*storageClassName, err).Error())
+
+				return &response
+			}
+		}
+	} else {
```
prometherion, line 63:
```suggestion
	if storageClassName != nil && if storageClassName != &allowed.Default {
		if csc, err = utils.GetStorageClassByName(ctx, c, *storageClassName); err != nil && !k8serrors.IsNotFound(err) {
			response := admission.Denied(NewStorageClassError(*storageClassName, err).Error())

			return &response
		}
	} else {
```

pkg/webhook/defaults/storage.go
```diff
@@ -0,0 +1,90 @@
...
+		mutate = true
+	}
+
+	if csc != nil && utils.IsDefaultStorageClass(csc) {
+		// If cluster default is not allowed in tenant
+		if !tnt.Spec.StorageClasses.MatchSelectByName(csc) {
+			mutate = true
+		}
+	}
```
prometherion, line 72:
```suggestion
	if csc != nil && utils.IsDefaultStorageClass(csc) && !tnt.Spec.StorageClasses.MatchSelectByName(csc) {
		// If cluster default is not allowed in tenant
		mutate = true
	}
```

pkg/webhook/defaults/storage.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if mutate {
+		pvc.Spec.StorageClassName = &tnt.Spec.StorageClasses.Default
+		recorder.Eventf(tnt, corev1.EventTypeNormal, "TenantDefault", "Assigned Tenant default Storage Class %s to %s/%s", allowed.Default, pvc.Namespace, pvc.Name)
```
prometherion, line 76: Let's record the error after error check occurring at line 81.

pkg/webhook/ingress/types.go
```diff
@@ -44,6 +47,24 @@ func (n NetworkingV1) IngressClass() (res *string) {
 	return
 }
 
+func (n NetworkingV1) SetIngressClass(ingressClassName string) {
+	if n.Spec.IngressClassName == nil {
+		if a := n.GetAnnotations(); a != nil {
+			if _, ok := a[annotationName]; ok {
```
prometherion, line 53: Do we need this check? `SetIngressClass` is called by the mutating webhook, accepting as its argument the default ingress class, we don't need to check if the object has already it.

pkg/webhook/ingress/types.go
```diff
@@ -44,6 +47,24 @@ func (n NetworkingV1) IngressClass() (res *string) {
...
+				return
+			}
+		}
+	}
+	// Assign in case the IngressClassName property was not set
+	if version, _ := utils.GetK8sVersion(); version.Major() == 1 && version.Minor() < 18 {
+		n.Annotations[annotationName] = ingressClassName
+	} else {
+		n.Spec.IngressClassName = &ingressClassName
+	}
```
prometherion, line 65: There's no need for this check since this API is served by the Kubernetes API Server that would reject any attempt against an unserver API group.

With that said, for the newer type we should avoid using the annotation that has been marked as deprecated: this strategy is for the API group networking.k8s.io/v1` and the field `IngressClassName` should be used.

pkg/webhook/utils/resources.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Get IngressClass by name (Does not return error if not found).
+func GetIngressClassByName(ctx context.Context, c client.Client, ingressClassName *string) (client.Object, error) {
+	if ingressClassName == nil {
+		return nil, nil
+	}
+
+	var obj client.Object
+
+	version, _ := GetK8sVersion()
```
prometherion, line 73: We're going to issue the Kubernetes version retrieval for each request: since this function is used in the file `pkg/webhook/defaults/handler.go`, please, initialize the retrieved version when calling `Handler`, considering that in `main.go` line 230 we already have the `kubeVersion` object, so you could inject it as a parameter.

pkg/webhook/utils/resources.go
```diff
@@ -0,0 +1,106 @@
...
+// Check if given IngressClass is cluster default.
+func IsDefaultIngressClass(class client.Object) bool {
+	annotation := "ingressclass.kubernetes.io/is-default-class"
+
+	if class != nil {
+		for a, v := range class.GetAnnotations() {
+			if a == annotation && v == TRUE {
+				return true
+			}
+		}
```
prometherion, line 102:
```suggestion
		annotations :=  range class.GetAnnotations() {
		if v, ok := annotations[annotation]; ok && v == TRUE {
			return true
		}
```

pkg/webhook/utils/resources.go
```diff
@@ -0,0 +1,106 @@
...
+// Check if given StorageClass is cluster default.
+func IsDefaultStorageClass(class client.Object) bool {
+	annotation := "storageclass.kubernetes.io/is-default-class"
+
+	if class != nil {
+		for a, v := range class.GetAnnotations() {
+			if a == annotation && v == TRUE {
+				return true
+			}
+		}
```
prometherion, line 59:
```suggestion
		annotations :=  range class.GetAnnotations() {
		if v, ok := annotations[annotation]; ok && v == TRUE {
			return true
		}
```
```

prometherion (changes requested):

pkg/webhook/ingress/types.go
prometherion: No, we don't.

prometherion (approved):

PR #684, opened 2023-01-23
Preventing  PersistentVolume cross-Tenant mount

### LINKED ISSUE #682: Support for PV labelling once used by a Tenant PVC
# Describe the feature

Labelling PVs with the Capsule Tenant label to allow retrieval using the capsule-proxy.

As a Tenant owner, I'd like to get my PV labelled as `capsule.clastix.io/tenant=MY_TENANT` in order to list them using the Capsule Proxy and attach it back.

# Expected behaviour

- Once a PV is created, a controller checks if it's used by a Tenant PVC, and if it's the case, it labels it with `capsule.clastix.io/tenant=MY_TENANT`
- A webhook should disallow attaching a PV to another PVC used by another Tenant to avoid data leak

prometherion:
Closes #682.

### DIFF
--- a/docs/content/general/tutorial.md
+++ b/docs/content/general/tutorial.md
@@ -2058,6 +2058,77 @@ Eventually, using the key `namespacedItem`, it is possible to reference existing
 
 As with `GlobalTenantResource`, the full reference of the API is available in the [CRDs API section](/docs/general/crds-apis).
 
+## Preventing PersistentVolume cross mounting across Tenants
+
+Any Tenant owner is able to create a `PersistentVolumeClaim` that, backed by a given _StorageClass_, will provide volumes for their applications.
+
+In most cases, once a `PersistentVolumeClaim` is deleted, the bounded `PersistentVolume` will be recycled due.
+
+However, in some scenarios, the `StorageClass` or the provisioned `PersistentVolume` itself could change the retention policy of the volume, keeping it available for recycling and being consumable for another Pod.
+
+In such a scenario, Capsule enforces the Volume mount only to the Namespaces belonging to the Tenant on which it's been consumed, by adding a label to the Volume as follows.
+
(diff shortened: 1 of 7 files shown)


docs/content/general/tutorial.md
```diff
@@ -2058,6 +2058,77 @@ Eventually, using the key `namespacedItem`, it is possible to reference existing
...
+  storageClassName: standard
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 3Gi
+  volumeName: pvc-9788f5e4-1114-419b-a830-74e7f9a33f5d
+
+$: k apply -f /tmp/pvc.yaml
+Error from server: error when creating "/tmp/pvc.yaml": admission webhook "pvc.capsule.clastix.io" denied the request: PeristentVolume pvc-9788f5e4-1114-419b-a830-74e7f9a33f5d cannot be used by the following Tenant, preventing a cross-tenant mount
```
radekg, line 2129: This error message is leaking the fact that there's a tenant who already claims this pvc, no?

prometherion: @radekg it depends, we're not reporting the Tenant name, rather, it's informing the Tenant Owner (so, the current Tenant) that it's in use.

Open in having a broader discussion, maybe in a new issue?

radekg: Sure, we can open a new issue. I’m just pointing it out after a brief look at the pr. There’s nothing _it depends_. Anybody who will execute this command will see that error. I understand that the tenant name is exposed but the error is already leaking a fact that this pvc is owned by someone else. How useful that information could be, difficult to say.

prometherion: Sure, looking forward to the issue to start the discussion!

Issue #680, opened 2023-01-09
feat: default pod security standards


oliverbaehler:
# Describe the feature

Currently we don't allow users to specify any pod security relevant properties (`securityContext` on pod and container level). By default we should deliver the capsule installation with all the necessary configuration, to run within a common pod security profile environment. Preferably, all components should be executable in a [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) profile.

# What would the new user story look like?

Add pod security and ingress options to make the default installation of capsule as secure as possible by default.

oliverbaehler:
Running the manager with the following securityContext (Container):

```
securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL
          readOnlyRootFilesystem: true
          runAsGroup: 1002
          runAsNonRoot: true
          runAsUser: 1002
```

Does not allow to run Capsule. It seems the `tls-controller` needs more privileges. The logs look like this:

```
{"level":"info","ts":"2023-01-09T11:11:46.239Z","logger":"setup","msg":"Capsule Version v0.1.3 b61db25.dev.dirty"}
{"level":"info","ts":"2023-01-09T11:11:46.239Z","logger":"setup","msg":"Build from: git@github.com:clastix/capsule.git"}
{"level":"info","ts":"2023-01-09T11:11:46.239Z","logger":"setup","msg":"Build date: 2023-01-05T23:52:03"}
{"level":"info","ts":"2023-01-09T11:11:46.239Z","logger":"setup","msg":"Go Version: go1.18.9"}
{"level":"info","ts":"2023-01-09T11:11:46.239Z","logger":"setup","msg":"Go OS/Arch: linux/amd64"}
{"level":"info","ts":"2023-01-09T11:11:46.347Z","logger":"controller-runtime.metrics","msg":"Metrics server is starting to listen","addr":":8080"}
{"level":"info","ts":"2023-01-09T11:11:46.444Z","logger":"controllers.TLS","msg":"Generating new TLS certificate"}
{"level":"error","ts":"2023-01-09T11:12:24.342Z","logger":"controllers.TLS","msg":"cannot update Capsule TLS","error":"client rate limiter Wait returned an error: context canceled","stacktrace":"main.main\n\t/workspace/main.go:192\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
[...]

PR #693, opened 2023-02-02
HELM: Add control for securityContext for capsule pods

zvlb:
Signed-off-by: Zemtsov Vladimir <vl.zemtsov@gmail.com>

Small feature for control securityContext in helm

### DIFF
--- a/charts/capsule/templates/daemonset.yaml
+++ b/charts/capsule/templates/daemonset.yaml
@@ -37,7 +37,7 @@ spec:
       {{- with .Values.securityContext }}
       securityContext:
         {{- toYaml . | nindent 8 }}
-      {{-- end }}
+      {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/capsule/templates/daemonset.yaml
+++ b/charts/capsule/templates/daemonset.yaml
@@ -34,6 +34,10 @@ spec:
       dnsPolicy: ClusterFirstWithHostNet
       {{- end }}
       priorityClassName: {{ .Values.priorityClassName }}
+      {{- with .Values.securityContext }}
+      securityContext:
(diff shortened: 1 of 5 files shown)


prometherion:
Hey @zvlb, thanks for the PR! I think this is strictly related to #680 by @oliverbaehler: would be great if we could add the defaults detected by Oliver, and let run the `ct` job in this PR that we have that ensures `capsule` can start and bind to the network.

charts/capsule/templates/daemonset.yaml
```diff
@@ -34,6 +34,10 @@ spec:
       dnsPolicy: ClusterFirstWithHostNet
       {{- end }}
       priorityClassName: {{ .Values.priorityClassName }}
+      {{- with .Values.securityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{-- end }}
```
prometherion, line 40: It should be `{{- end }}`

prometherion (changes requested):
Please, adhere to the semantic commit messages, more info is available [here](https://www.conventionalcommits.org/en/v1.0.0/).

For Helm changes, I use `chore(helm)` as the subject, and remember to bump up the patch version in the `Chart.yaml` file so we can release a new Chart version.

Also, the Chart documentation must be aligned to the new changes, please, issue `make helm-docs` and commit the result.

zvlb:
Oliver has errors with securityContext, like:
`
securityContext:
          capabilities:
            drop:
            - ALL
          readOnlyRootFilesystem: true
          runAsGroup: 1002
          runAsNonRoot: true
          runAsUser: 1002
`

Maybe now we will deploy a new chart version with control securityContext
Today or tomorrow I will test best practice for securityContext, and set PR with default values

prometherion:
May I ask you to add the required `securityContext` suggested by Oliver, along with your proposed fix?

zvlb:
I reopen this PR soon

PR #702, opened 2023-02-10
feat: add runtimeclass kind

oliverbaehler:
Adds the RuntimeClass field to the owner API. I think this change should be possible without any problems. Required for https://github.com/clastix/capsule-proxy/issues/260.

### DIFF
--- a/api/v1beta2/owner.go
+++ b/api/v1beta2/owner.go
@@ -34,7 +34,7 @@ func (p ProxyOperation) String() string {
 	return string(p)
 }
 
-// +kubebuilder:validation:Enum=Nodes;StorageClasses;IngressClasses;PriorityClasses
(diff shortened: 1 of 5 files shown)


prometherion (approved):

prometherion (changes requested):
We need to update the `kubebuilder` enum `ProxyServiceKind` too, otherwise the OAPI server will reject the request due to a validation error:

```go
// +kubebuilder:validation:Enum=Nodes;StorageClasses;IngressClasses;PriorityClasses;RuntimeClasses
```

The problem is that we need an update of the CRD that is required manually, therefore is a breaking change and requires a minor release.

prometherion:
Closing in favour of a new MR.

PR #704, opened 2023-02-14
feat(helm): add securityContexts to jobs

zvlb:
local test:

```bash
~ kind create cluster
...

~ k cluster-info
Kubernetes control plane is running at https://127.0.0.1:61015
CoreDNS is running at https://127.0.0.1:61015/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.

~ helm install capsule . -n capsule-system --create-namespace 
NAME: capsule
LAST DEPLOYED: Tue Feb 14 17:34:56 2023
NAMESPACE: capsule-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
...

~ k create -f - << EOF
apiVersion: capsule.clastix.io/v1beta2
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - name: alice
    kind: User
EOF
tenant.capsule.clastix.io/oil created

~ k get tenants
NAME   STATE    NAMESPACE QUOTA   NAMESPACE COUNT   NODE SELECTOR   AGE
oil    Active                     0                                 6s

~ ../../hack/create-user.sh alice oil
creating certs in TMPDIR /var/folders/c7/78_t4pvj5zd2qrcwgq05h7_40000gn/T/tmp.EC4Ogyzl
merging groups /O=capsule.clastix.io
Generating RSA private key, 2048 bit long modulus
......................................................................+++
.+++
e is 65537 (0x10001)
certificatesigningrequest.certificates.k8s.io/alice-oil created
certificatesigningrequest.certificates.k8s.io/alice-oil approved
kubeconfig file is: alice-oil.kubeconfig
to use it as alice export KUBECONFIG=alice-oil.kubeconfig

~ export KUBECONFIG=alice-oil.kubeconfig
~ k create namespace oil-development
namespace/oil-development created

~ export KUBECONFIG=
~ k get tenant oil
NAME   STATE    NAMESPACE QUOTA   NAMESPACE COUNT   NODE SELECTOR   AGE
oil    Active                     1                                 31s
```

PR #701, opened 2023-02-10
refactor(helm): add control for securityContext

zvlb:
Signed-off-by: Zemtsov Vladimir <zemtsov.v@mail366.com>

Small feature for control securityContext in helm

Local test:
```bash
~ kind create cluster
...


~ k cluster-info
Kubernetes control plane is running at https://127.0.0.1:60977
CoreDNS is running at https://127.0.0.1:60977/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.


~ helm install capsule . -n capsule-system --create-namespace
NAME: capsule
LAST DEPLOYED: Fri Feb 10 09:57:02 2023
NAMESPACE: capsule-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
  ...

~ k get deployments.apps -n capsule-system capsule-controller-manager -o yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  ...
  name: capsule-controller-manager
  namespace: capsule-system
  ...
spec:
  ...
  template:
    ...
    spec:
      containers:
      - ...
        ...
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL
          readOnlyRootFilesystem: true
        ...
      ...
      securityContext:
        runAsGroup: 1002
        runAsNonRoot: true
        runAsUser: 1002
      ...



~ k create -f - << EOF
apiVersion: capsule.clastix.io/v1beta2
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - name: alice
    kind: User
EOF
tenant.capsule.clastix.io/oil created


~ k get tenants
NAME   STATE    NAMESPACE QUOTA   NAMESPACE COUNT   NODE SELECTOR   AGE
oil    Active                     0                                 4s


~ ../../hack/create-user.sh alice oil
creating certs in TMPDIR /var/folders/c7/78_t4pvj5zd2qrcwgq05h7_40000gn/T/tmp.Sr8bOiQ3
merging groups /O=capsule.clastix.io
Generating RSA private key, 2048 bit long modulus
......................................+++
........+++
e is 65537 (0x10001)
certificatesigningrequest.certificates.k8s.io/alice-oil created
certificatesigningrequest.certificates.k8s.io/alice-oil approved
kubeconfig file is: alice-oil.kubeconfig
to use it as alice export KUBECONFIG=alice-oil.kubeconfig


[...]

PR #705, opened 2023-02-15
feat: add bash script for local-test capsule

zvlb:
Hi.

In the future, I will actively use capsule in my K8s clusters. For local test before PR in [this](https://github.com/clastix/capsule/pull/704) and [this](https://github.com/clastix/capsule/pull/701) PR I use kind.

I think this script will speed up local testing in the future


prometherion (approved):
LGTM, thanks for the contribution!

PR #717, opened 2023-02-28
feat: allow customization of the webhook port

### LINKED ISSUE #715: Port 9443 conflict
We see that capsule-webhook-service is using port 9443 which conflicts with other running services in the cluster. We tried changing the port in the webhook-service.yaml but it is throwing error. We are able to install Capsule only when the port is 9443 so it looks like the port is hard-coded in the Capsule code.  

Tested on current and previous version 
chart:  capsule-0.3.5 
app version: 0.2.1

1) How do we change the port 9443 for webhook-service? Can you please provide the details?
2) Is it possible to provide port as configurable item in Capsule installation?

aslafy-z:
Fixes #715

In some networking contexts, operators may need to use a custom webhook port. This PR add the `webhook-port` CLI parameter, sets its default value to `9443` and passes it down to the manager. It also adds the `manager.webhookPort` Helm value to the chart and bindings to the service, deployment and daemonset resources.

### DIFF
--- a/charts/capsule/Chart.yaml
+++ b/charts/capsule/Chart.yaml
@@ -21,7 +21,7 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.3.6
+version: 0.4.0
 
 # This is the version number of the application being deployed.
 # This version number should be incremented each time you make changes to the application.
--- a/charts/capsule/README.md
+++ b/charts/capsule/README.md
@@ -107,6 +107,7 @@ Here the values you can override:
 | manager.resources.limits.memory | string | `"128Mi"` |  |
 | manager.resources.requests.cpu | string | `"200m"` |  |
 | manager.resources.requests.memory | string | `"128Mi"` |  |
+| manager.webhookPort | int | `9443` | Set an alternative to the default container port.  Useful for use in some kubernetes clusters (such as GKE Private) with aggregator routing turned on, because pod ports have to be opened manually on the firewall side |
 
 ### ServiceMonitor Parameters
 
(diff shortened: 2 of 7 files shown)


prometherion (changes requested):
Outstanding contribution, @aslafy-z!

May I ask you to split the changes into two commits? The first one with the code-base changes, the second with Helm ones, and a minor bump app of the Chart version.

The reason behind this is to decouple the Helm Chart from the code-base, so we can release both independently.

prometherion:
Please, along with the Helm changes, issue `make helm-docs` to generate the documentation.

aslafy-z:
Sure, I will make the changes! Do you want me to expose the webhookHost as well ?

aslafy-z:
@prometherion The appVersion minor will be bumped automatically, or do I need to do it?

charts/capsule/Chart.yaml
```diff
@@ -21,7 +21,7 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.3.5
+version: 0.4.0
```
prometherion, line 24: My personal experience: adding values doesn't require a minor bump, a patch is good enough.

We're using minor changes when updating the CRD definition, and major when changing or deprecating Helm values.

Please, let's coordinate with #718 to get it merged, and then bump up to the next patch version.

prometherion (changes requested):
Regarding the `appVersion` I was expecting to release it for v0.2.3, however, we accumulated some changes that must be addressed with v0.3.0.

Let's go for this, I'll publish the newer Chart once v0.3.0 is ready.

prometherion:
@aslafy-z I just contributed to your branch by fixing the conflict, going to merge this to proceed to tag Capsule v0.3.0 with several fixes and this enhancement.

prometherion:
Helm testing is failing due to v0.2.2 not being released on the main branch, going to merge this and fix this once v0.3.0 is cut off.

prometherion (approved):

PR #718, opened 2023-03-01
chore(helm): bump up the appVersion

sagar-jadhav:
The latest tag in capsule is `0.2.2` but the appVersion in the chart.yaml is still pointing to the previous release i.e. `0.2.1` so incremented the appVersion.

### DIFF
--- a/charts/capsule/Chart.yaml
+++ b/charts/capsule/Chart.yaml
@@ -21,8 +21,8 @@ sources:
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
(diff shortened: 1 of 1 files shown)


prometherion (changes requested):
May I ask you to bump up also the Chart so we can release it?

sagar-jadhav:
> May I ask you to bump up also the Chart so we can release it?

@prometherion I have bumped up the chart version as suggested.

prometherion (approved):
`golangci-lint` is failing, although there's no changes in the code-base /shrug

Thanks for the contribution, merging!

PR #728, opened 2023-03-07
Adding Fastweb as adopter

gabrilabs75:
Adding Fastweb as adopter


prometherion (approved):
LGTM and thanks a lot, I'm going to fix the commit message to adhere to the semantic rule!

Issue #733, opened 2023-03-22
Unable to patch the namespace not managed by Capsule


sagar-jadhav:
# Bug description

As a capsule user, I am unable to patch the namespace not managed by capsule, even though I have explicit access to patch that namespace.

# How to reproduce

1. Create a namespace called `oil-sj` as a cluster admin
2. Follow the https://capsule.clastix.io/docs/general/getting-started/ tutorial
3. As an cluster admin create the following ClusterRole and ClusterRoleBinding
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: alice-ns-provisioner
rules:
- apiGroups:
  - ""
  resources:
  - namespaces
  resourceNames:
  - oil-sj
  verbs:
  - create
  - patch
  - get
``` 
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: alice-ns-provisioner
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: alice-ns-provisioner
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: alice
```
4. Login to cluster using alice kubeconfig and run the `kubectl annotate ns oil-sj env=stg` command. Command fails with below error
```
Error from server: admission webhook "owner.namespace.capsule.clastix.io" denied the request: the OwnerReference cannot be removed
```
# Expected behavior

As a capsule user I should be able to patch the namespace not managed by capsule, If I have explicit access to patch that namespace.

# Logs

N/A

# Additional context

- Helm Chart version: capsule-0.3.1

prometherion:
Planning to release this for v0.3.1, let's wait some days so I can work also on #732.

sagar-jadhav:
> Planning to release this for v0.3.1, let's wait some days so I can work also on #732.

Sure. Thanks

PR #734, opened 2023-03-22
fix: fixed namespace webhooks

sagar-jadhav:
Fixes https://github.com/clastix/capsule/issues/733

### DIFF
--- a/pkg/webhook/namespace/owner_reference.go
+++ b/pkg/webhook/namespace/owner_reference.go
@@ -42,6 +42,10 @@ func (r *ownerReferenceHandler) OnUpdate(_ client.Client, decoder *admission.Dec
 			return utils.ErroredResponse(err)
 		}
 
+		if len(oldNs.OwnerReferences) == 0 {
+			return nil
+		}
+
 		newNs := &corev1.Namespace{}
 		if err := decoder.Decode(req, newNs); err != nil {
 			return utils.ErroredResponse(err)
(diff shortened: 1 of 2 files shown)


pkg/webhook/namespace/patch.go
```diff
@@ -72,11 +72,6 @@ func (r *patchHandler) OnUpdate(c client.Client, decoder *admission.Decoder, rec
 
 				return &response
 			}
-		} else {
-			recorder.Eventf(ns, corev1.EventTypeWarning, "NamespacePatch", e)
-			response := admission.Denied(e)
-
-			return &response
```
prometherion, line 79: I'm a bit concerned about removing this, may I ask you for some support in getting the reason?

pkg/webhook/namespace/owner_reference.go
```diff
@@ -42,6 +42,10 @@ func (r *ownerReferenceHandler) OnUpdate(_ client.Client, decoder *admission.Dec
 			return utils.ErroredResponse(err)
 		}
 
+		if len(oldNs.OwnerReferences) == 0 {
+			return nil
+		}
+
```
prometherion, line 48: This is a nice catch!

pkg/webhook/namespace/patch.go
rakshpatel: @prometherion As @sagar-jadhav stated above non-capsule managed namespaces will not have this label, in that case, this else will always block. Is there any use case for which this is needed?

prometherion: You're right, thanks for pointing it out!

PR #735, opened 2023-03-22
chore(adopters): adding Velocity as adopter

yarelm:

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -10,5 +10,8 @@ This is a list of companies that have adopted Capsule, feel free to open a Pull-
 ### [Fastweb](https://www.fastweb.it/)
 ![Fastweb](https://www.fastweb.it/grandi-aziende/gfx/common/logo-fastweb-header.svg)
 
+### [Velocity](https://velocity.tech/)
+![Velocity](https://raw.githubusercontent.com/yarelm/velocity-logo/main/Light.jpg)
+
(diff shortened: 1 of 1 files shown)


ADOPTERS.md
```diff
@@ -10,5 +10,8 @@ This is a list of companies that have adopted Capsule, feel free to open a Pull-
 ### [Fastweb](https://www.fastweb.it/)
 ![Fastweb](https://www.fastweb.it/grandi-aziende/gfx/common/logo-fastweb-header.svg)
 
+### [Velocity](https://velocity.tech/)
+![Velocity](https://raw.githubusercontent.com/yarelm/velocity-logo/main/Light.jpg)
```
prometherion, line 14: Couldn't we use [this](https://uploads-ssl.webflow.com/6361517a3f02a635b3a4220b/6377dc818825713bdcc40663_velocity-logo.png) image, rather?

yarelm: We can, but I prefer to decouple it from our website since we may do a future remodeling of it and forget about this usage.

But I don't mind if you prefer it that way. :)

prometherion: The image is a bit big, that's the point, it would be great if you could upload the resized one in your repository, or, add a style in MD (not sure if it's possible) to set a width.

yarelm: Oh I missed your point.

I changed the image now to the resized one.

PR #736, opened 2023-03-23
Blocking replicated resources write operations put in place by tenant owners

### LINKED ISSUE #732: Prevent replicated resources by TenantResource to get deleted by the Tenant Owner
A validating webhook should block the deletion of the replicated resources by `GlobalTenantResource` and `TenantResource` in order to avoid waiting for the next reconciliation operation.

prometherion:
Closes #732.

@MaxFedotov @oliverbaehler @bsctl WDYT of this? Essentially, replicated resources managed through a `GlobalTenantResource` or a `TenantResource` cannot be deleted or edited by Tenant Owners. This leads to avoiding an inconsistent state of the replicated resources.

Keep in mind that resources can be deleted by admin users or k8s controller managers.


bsctl:
@prometherion that's looks good. Just suggest to use the same strategy for all types of tenant resources, i.e. blocking vs reconciliating. Do we have different ways to handle tenant resources in the current implementation?

prometherion:
> Just suggest to use the same strategy for all types of tenant resources

We can discuss it in a new issue/discussion for the upcoming release, but definitely it would be interesting, although breaking from the API perspective because, if I understood correctly, it means moving the NetworkPolicy and LimitRange definitions from the Tenant API to the GlobalTenantResource.

pkg/webhook/tenantresource/objects.go
```diff
@@ -0,0 +1,77 @@
...
+	obj := &unstructured.Unstructured{}
+	if err := decoder.Decode(req, obj); err != nil {
+		return utils.ErroredResponse(err)
+	}
+	// If the object is not managed by a TenantResource thanks to its label, ca ne skipped
+	if labels := obj.GetLabels(); labels != nil {
+		if _, ok := labels[resources.Label]; !ok {
+			return nil
+		}
+	}
```
prometherion, line 46: I'm still not convinced by this approach: I'd like to give it a try to the indexer since otherwise a Tenant Owner could lock a resource object by just using the label key.

PR #753, opened 2023-04-20
fix: missing v1beta2 version for CRD ConversionReviewVersions

### LINKED ISSUE #752: Missing ConversionReviewVersions in updateTenantCustomResourceDefinition function
# Bug description
In helm chart both `tenant` CRD
https://github.com/clastix/capsule/blob/953cfdc172dd1dd93646dd8609cda61071c8528a/charts/capsule/crds/tenant-crd.yaml#L17-L20)
and `capsuleConfiguration` CRD
https://github.com/clastix/capsule/blob/953cfdc172dd1dd93646dd8609cda61071c8528a/charts/capsule/crds/capsuleconfiguration-crd.yaml#L18-L21

specifies `v1beta2` in `conversionReviewVersions`.

But in `tlsReconciler` this version is missing:

https://github.com/clastix/capsule/blob/953cfdc172dd1dd93646dd8609cda61071c8528a/controllers/tls/manager.go#L243

MaxFedotov:
closes #752

### DIFF
--- a/controllers/tls/manager.go
+++ b/controllers/tls/manager.go
@@ -240,7 +240,7 @@ func (r *Reconciler) updateTenantCustomResourceDefinition(ctx context.Context, n
 						},
 						CABundle: caBundle,
 					},
-					ConversionReviewVersions: []string{"v1alpha1", "v1beta1"},
+					ConversionReviewVersions: []string{"v1alpha1", "v1beta1", "v1beta2"},
 				},
 			}
 


prometherion (approved):
Thanks for spotting this, Max!

May I ask you to provide in the issue the downside of this bug to inform the end-users what they're gonna face if not upgrading to the latest version?

That will be gold for the changelog when releasing v0.3.2

PR #759, opened 2023-05-10
fix: update golang to 1.19

slimm609:
update capsule to golang 1.19 to allow for easier recompile using boringcrypto which was added to main in golang 1.19

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.18'
+          go-version: '1.19'
       - run: make installer
       - name: Checking if YAML installer file is not aligned
         run: if [[ $(git diff | wc -l) -gt 0 ]]; then echo ">>> Untracked generated files have not been committed" && git --no-pager diff && exit 1; fi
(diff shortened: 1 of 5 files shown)


prometherion (approved):

prometherion:
Hey @slimm609, thanks for the fix!

It would be great to split the changes into three separate commits if you don't mind, something as:

```
chore(build): golang 1.19 update
chore(gh): golang 1.19 update
docs: golang 1.19 requirement
```

We're following the semantic commit messages guidelines

slimm609:
@prometherion what is the reason for multiple PRs?   They are direct dependencies of each other.   That would be like saying split the update of library A and Library B into 2 PRs even though they directly depend on each other.  If the CI is 1.19 and the image is 1.18, that's not a valid CI test as its not an accurate test of what the build would produce.

meetdpv:
When are we expecting to release v0.3.2 to be available with the updated docker images?

prometherion:
@meetdpv once we have the [milestone](https://github.com/clastix/capsule/milestone/10) completed, I'll cut off a new version.

meetdpv:
Thanks. That's great to hear. Could you please let us know the estimated timeline because if it is pushed to a month, we won't be able to whitelist Capsule image from security team in order to take it to Prod?

PR #782, opened 2023-06-27
fix: remove ownerReferences from tenantResource namespacedItems before applying them

### LINKED ISSUE #781: Capsule should drop ownerReference for {Global}TenantResource namespacedItems
# Bug description

Assume a user wants to share a TLS secret with a certificate, which was created by `cert-manager`. This secret will obviously have `ownerReference` to a `certificate` object. When Capsule will copy this secret, it won’t drop this `ownerReference`, so the secret will be immediately deleted by gc.

# Expected behavior

Capsule should drop `ownerReference` for `namespacedItems` before creation.

MaxFedotov:
closes #781


prometherion (approved):
LGTM, let's see if the test suite is 🟢

PR #796, opened 2023-08-07
WIP: feat: managed mutation webhook

oliverbaehler:
This webhook adds a managed label to each resource which is created or updated in a capsule namespace. This is required for the capsule-proxy to query any namespaced resource.

 Helps with issues:

  * https://github.com/clastix/capsule-proxy/issues/303
  * https://github.com/clastix/capsule-proxy/issues/185

I have added an enabled switch for the webhook in the chart, since it brings quiet some workload and thee might be users, which don't use the capsule-proxy at all.

What this implementation does not respect, is the handling of existing resources. I would probably just add a simple script to the documentation, which adds the label to all existing resources in all tenant namespaces.


oliverbaehler:
Implementation is moved entirely to capsule-proxy.

Issue #803, opened 2023-08-23
Documentation Spelling Mistakes - v2


whatnick:
# Bug description

Like #197 I would like to fix some more spelling in docs that have crept in the last couple of years. Samples from reading [GitOps docs](https://capsule.clastix.io/docs/guides/flux2-capsule#full-setup) are:

`To have a glimp`
instead of
`To have a glimpse`

# How to reproduce

1. Read the docs in full
2. Pay attention to things outside YAML code blocks

# Expected behavior

Slightly better documentation quality

whatnick:
@prometherion I am spending some time today studying the Capsule docs and creating a `capsule.txt` list for a spellchecker automation which we could potentially run in CI here to solve the issue longer term. Would something like this in CI be useful ? https://github.com/marketplace/actions/check-spelling

prometherion:
Closes via #804.

PR #804, opened 2023-08-24
Add dictionary and fix some spellings in docs

whatnick:
Addresses #803 , approach taken was : 
- Capture possible misspellings list using [Typescript spellchecker](https://github.com/tbroadley/spellchecker-cli/issues)
- Weed out the obvious technology words identified and keep them
- Correct some spellings while acknowledging my bias towards Australian and British English

In the future I propose maintaining the `dictionary.txt` in the docs folder and automating via CI using the Typescript tool I used or some other pre-baked GitHub Action. I can create a follow-on issue with this if you would like some assistance with the CI.


whatnick:
Which CI workflow should I target ? A new one or a linting focused one
seems appropriate. Will propose `make docs-lint` as a command where things
like spellcheck and markdown lint can live.

prometherion:
`make docs-lint` LGTM.

We could set up a new workflow in order to trigger it only if we have changes on `docs` folder, as we're doing similarly with e2e.

https://github.com/clastix/capsule/blob/ffa4a47e6baa6f9cb3660e2f37b81392a40cf675/.github/workflows/e2e.yml#L3-L27

Makefile
```diff
@@ -275,3 +275,7 @@ e2e-exec: ginkgo
 
 e2e-destroy:
 	kind delete cluster --name capsule
+
+docs-lint:
+	cd docs/content && spellchecker *.md */*.md -d dictionary.txt
```
aslafy-z, line 280: Maybe at least use `npx spellchecker` instead?

aslafy-z: Also relying into the shell globbing is not ideal, simple quote the glob pattern to use the node implementation instead.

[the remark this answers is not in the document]
whatnick: Unfortunately no pre-built image.

whatnick: Made suggested changes.
- Disable global install in workflow
- Execute `spellchecker-cli` with npx in Makefile without global install
- Pass files to tool with `-f` flag + quoted globs to bypass shell globbing and activate tool internal node globbing.

.github/workflows/docs-lint.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  spelling:
+    name: Spell Check
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
```
aslafy-z, line 23: Depending on the size of the spellcheck-cli bundle, we might want to enable caching, if it's not done by default. In this case, I'm not sure whether npx targets the right directory if the package is not there. However, if the package is already installed, I think it uses it instead of fetching it again. So we might need to keep the npm install in the GitHub Action. To be tested

whatnick: I have tested by uninstalling `spellcheck-cli` and running npx. I looked like npx grabbed it into a temp spot and ran it. Did not appear in my global modules list after, but ran faster on future executions. Perhaps a different cache location than the standard global node modules cache.
I can perform these enhancements once GHA runs and we observe the behaviour :
- Re-enable npm install
- Activate GHA node-modules cache with `cache: 'npm'`

aslafy-z (approved):

PR #806, opened 2023-08-30
chore(grafana): configurable datastore

h4wkmoon:
In grafana dashboard, prometheus datasource is hardcoded. This does fit every installation.
This PR make prometheus datasource a templating variable.

### DIFF
--- a/config/grafana/dashboard.json
+++ b/config/grafana/dashboard.json
@@ -80,7 +80,7 @@
           "mode": "spectrum"
         },
         "dataFormat": "timeseries",
-        "datasource": "prometheus",
+        "datasource": "$DS_PROMETHEUS",
         "description": "",
         "fieldConfig": {
           "defaults": {},
@@ -155,7 +155,7 @@
         "type": "row"
       },
       {
-        "datasource": "prometheus",
+        "datasource": "$DS_PROMETHEUS",
         "fieldConfig": {
           "defaults": {
             "color": {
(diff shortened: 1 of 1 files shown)


prometherion (approved):

prometherion:
Lgtm, let's align to the semantic commit messages to get it merged!

I'd say `chore(grafana): configurable datastore` would be good.

config/grafana/dashboard.json
```diff
@@ -1999,10 +1999,27 @@
     "tags": [],
     "templating": {
       "list": [
+      {
+          "current": {
+            "selected": false,
+            "text": "Prometheus",
+            "value": "Prometheus"
+          },
+          "hide": 0,
```
aslafy-z, line 2008:
```suggestion
          "hide": 2,
```
So the variable is hidden and the default Prometheus data source is selected

aslafy-z (commented):
I generally see the `${DS_PROMETHEUS}` notation but I guess it's ok without brackets

config/grafana/dashboard.json
h4wkmoon: I'd think it's better to have this variable visible. For example, one may want to switch between prometheus (with short term storage) and Thanos.

h4wkmoon:
> I generally see the `${DS_PROMETHEUS}` notation but I guess it's ok without brackets

True. Let me check.

config/grafana/dashboard.json
aslafy-z: Ok to me. :)

prometherion:
@h4wkmoon can we convert this from draft to ready for getting merged?

oliverbaehler (approved):

Issue #824, opened 2023-10-14
[UMBRELLA] Publish supply chain security artifacts during the release


maxgio92:
# Describe the feature

This is a security feature on the distribution of the Capsule OCI image artifact.
Security artifacts can help secure the supply chain by providing ways for consumers to verify the authenticity of the artifacts provided by the project, the integrity of the artifacts.

# What would the new user story look like?

As an end-user of Capsule, I'd like to verify the authenticity and integrity of artifacts officially provided by the Capsule project.

# Expected behavior

* [ ] Provide OCI image's signature
* [ ] Provide SBOM of the artifacts
* [ ] Provide provenance attestation
* [ ] TBD

maxgio92:
Hey @oliverbaehler, can this issue be considered completed?
Thanks!

prometherion:
I'd say so, let's wait for Oliver's feedback.

oliverbaehler:
Implemented, except SBOM for charts

PR #825, opened 2023-10-15
feat(capsule): release security and workflow updates

oliverbaehler:
# Description

implements #824 
related to #820 

This quiet large Pull Request has several improvements to the capsule repository. They are both related to the move of the project but are mainly focused on the Github actions part. The release flows are kept as is (tag = image release, helm-v* = helm chart release).

Note before review:

  * the helm-test workflow won't execute successful, because it already tries to pull the image from the new location. Which will only work after the first release.
 
## Release

On tag a release is created which publishes the controller binary. The Release is created automatically and adds a changelog based on the commits added.

With the commit check, we verify the types of changes. The changelog respects the types and creates a Changelog via goreleaser. Additional documentation has been added as part of CONTRIBUTING.md.

## Packages

With the introduction of these workflows the following packages will be published:

* `capsule` - Capsule Controller Docker Image
* `charts/capsule` - Capsule OCI Helm Chart
* `signatures` - Signature Registry for Docker Images and OCI Helm Charts
* `sbom` - SBOM Registry

You can see as reference the fork's package repository:
https://github.com/orgs/buttahtoast/packages?repo_name=capsule

## Docker Image

Changed to build workflow to use [ko](https://github.com/ko-build/ko). Since we don't need to release any binaries, I did not consider goreleaser. Ko builds docker images automatically and requires minimal configuration in the `.ko.yaml` file. Currently the controller is released for  `linux/arm64` and `linux/amd64`. Most of the changes in the Makefile relate to this change. I have made sure the ko build works locally and can be used with local e2e testing etc. 

We have two workflows, a docker-build workflow on pull requests. Which builds the image and ensures everything works as expected. The docker publish action publishes the docker image to `capsule`.  Ko automatically generates an sbom for the oci. My github action signs the published sigest and publishes the signature under `signatures`. At the end an attestation for the image is made via SLSA and uploaded to the `capsule` package repo.

Important: Ko does not require a Dockerfile. The entrypoint is `/ko-build/capsule`. Therefor the `manager` command is no longer required/supported.

## Helm Chart
I have kept the old release procedure to publish a `tar.gz` to a dedicated `charts` repository (https://github.com/projectcapsule/charts). But this workflow is the least tested, it might be, that we need additional changes. But I would add them, should it fail.

I have focused on adding a workflow to publish the capsule helm chart in oci format. This happens along side with the legacy-release. The OCI helm chart is also signed and we also provide attestation for the build via slsa.

The chart documentation has been updated to point to the new repository and includes instructions on OCI usage. In addition the helm-test actions have been reneabled.

## Github-Workflow Security

Introduces a new github-check (`.github/workflows/check-actions.yaml`), which requires all github actions to be included be their commit SHA. In addition a matching dependabot configuration has been added, which updates github actions automatically. I have initially added all the latest commit SHA for all the Github actions used.

## Code Coverage

Enabled Code Coverage on the upstream repository. Added a workflow to upload coverage on pull requests. Added the badge to the `README.md`

## ScoreBoard

Added periodical ScoreBoard action.

I think these cover most of the changes. There will be some cleanup work for me after that. But with these workflows we should satisfy CLO monitor and provide secure artifacts to our users.


charts/capsule/templates/deployment.yaml
```diff
@@ -60,8 +60,6 @@ spec:
             secretName: {{ include "capsule.secretTlsName" . }}
       containers:
         - name: manager
-          command:
```
prometherion, line 63: May I ask you the reason in removing the command, here?

oliverbaehler: with ko, we no longer build a binary which is called `manager`. We would have to replace it with `/ko-build/capsule`. I thought it's nicer to just remove it.

oliverbaehler:
@prometherion i would like to get this merged. Its far from perfect but i would like to make additional adjustments with follow-up pull requests

prometherion:
Sure thing, may I just ask you to squash where possible the commits to make easier tracking the changes, please?

oliverbaehler:
@prometherion done

prometherion (approved):
Terrific job, LGTM! 🚀

Issue #812, opened 2023-09-23
[UMBRELLA] CNCF Sandbox project onboarding 


maxgio92:
**General**

This is an umbrella issue to track the status of all the required points that need to be addressed in the Git repository and GitHub:

- [x] Review and expand the open governance. See [CNCF governance](https://contribute.cncf.io/maintainers/governance/) - cncf/toc#901 
- [x] Move the GitHub repository in its own separate neutral GitHub organization (ensure 'thelinuxfoundation' and 'caniszczyk' are added as initial org owners)
- [ ] Move documentation to MkDocs - cncf/toc#821 
- [x] Ensure [DCO](https://github.com/apps/dco) or [CLA](https://github.com/cncf/cla) are enabled for all GitHub repositories of the project
- [x] Ensure that hat the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md) (or your adopted version of it) are explicitly referenced at the project's README on GitHub
- [x] CII: Start on a CII best practices badge https://bestpractices.coreinfrastructure.org/en
- [x] Adopt a license scanning tool, like [FOSSA](https://fossa.com/) or [Snyk](https://snyk.io/)

**Additional informations**

Related to: https://github.com/cncf/sandbox/issues/166

prometherion:
License scanning has been introduced with #810 by @ptx96

prometherion:
CII has been addressed with #816.

prometherion:
Adhering to CNCF ToC is addressed with #817.

oliverbaehler:
@prometherion @maxgio92 @MaxFedotov  Do we also need to create a new organisation on artifacthub? I would like to start with trivial stuff in the CLO Monitor like Artifacthub Badges etc. Should I create an organisation and add all the maintiners?

prometherion:
It makes definitely sense.

I just created the organization on Artifact Hub, please, share your usernames (publicly, or privately) so I can add you there: thanks for the heads up, @oliverbaehler! ❤️

MaxFedotov:
> It makes definitely sense.
> 
> I just created the organization on Artifact Hub, please, share your usernames (publicly, or privately) so I can add you there: thanks for the heads up, @oliverbaehler! ❤️

Hey! Mine is the same as here, MaxFedotov

oliverbaehler:
@prometherion should we still make changes against the current capsule repostories or should we qait until its moved to the new organisation?

[...]

PR #827, opened 2023-10-16
Moving to projectcapsule/capsule

prometherion:
Required for CNCF Sandboxing, related to #812.


oliverbaehler:
@prometherion looks good at first sight. Somehting, each file has the following license header:

```
// Copyright 2020-2021 Clastix Labs
// SPDX-License-Identifier: Apache-2.0
```

This must be changed I would say. We need to merge this or either my merge request and then do a final cleanup :D, wont chatch them all in one

prometherion:
Thanks @oliverbaehler, last commit addressed it.

Definitely, the migration would require some reworks since we cannot catch everything.

Let's start the migration so we can easily solve all of them, but at least, we would make possible to other projects to import properly the source code.

PR #845, opened 2023-10-23
feat: improve changelog documentation and linting

oliverbaehler:
Adds a lot of documentation, references changelog to where it's created. Ensures commits are formatted to or changelog's standards. The biggest dicussion point here, is regarding the types of changes and their scope which we accept. Most of the other changes are based on the ruleset. 

Let me know what you think: 

https://github.com/buttahtoast/capsule-cncf-slam/blob/feat/doc-changelog/CONTRIBUTING.md#semantics


prometherion (approved):
Terrific, LGTM! 🚀

Issue #737, opened 2023-03-24
Adding podOptions to the Tenant specification


prometherion:
The current `v1beta2` Tenant API resource has multiple specs that could be added in a single struct.

```yaml
apiVersion: capsule.clastix.io/v1beta2
kind: Tenant
metadata:
  name: test
spec:
  //  Specifies the allowed RuntimeClasses assigned to the Tenant. Capsule
  // assures that all Pods resources created in the Tenant can use only one of
  // the allowed RuntimeClasses. Optional.
  runtimeClasses:
  // Specifies the allowed priorityClasses assigned to the Tenant. Capsule
  // assures that all Pods resources created in the Tenant can use only one of
  // the allowed PriorityClasses. A default value can be specified, and all the
  // Pod resources created will inherit the declared class. Optional.
  priorityClasses:
  // Specify the allowed values for the imagePullPolicies option in Pod                                                                                                                                            
  // resources. Capsule assures that all Pod resources created in the Tenant can                                                                                                                                   
  // use only one of the allowed policy. Optional.
  imagePullPolicies:
  // Specifies the trusted Image Registries assigned to the Tenant. Capsule                                                                                                                                        
  // assures that all Pods resources created in the Tenant can use only one of                                                                                                                                     
  // the allowed trusted registries. Optional.
  containerRegistries:
```

The proposal is to group all of these in a new spec parent key, such as `podOptions`, as we're doing with `namespaceOptions`, `ingressOptions` and `serviceOptions`.

```yaml
apiVersion: capsule.clastix.io/v1
kind: Tenant
metadata:
  name: test
spec:
  podOptions:
    runtimeClasses:
    priorityClasses:
    imagePullPolicies:
    containerRegistries:
```

@bsctl @oliverbaehler @MaxFedotov please, share your thoughts!

oliverbaehler:
@prometherion makes sense to me. But since this is again an API bump i would like to explore more generic options as well. 

## Moving towards Kyverno Policies

This is just a super experimental thought regarding this topic. Just wanted to hear you guys thoughts about it. If it could be remotely interesting we should move it to.a new issues. This would require a rewrite 

After some thinking I also thought, if we would want to reconsider moving the capsule policy engine towards Kyverno policies. Kyverno seems to be the most adapted policy engine by now and we have already seen so many use-cases where it's in use. I am coming up with that idea, since it would change .



So just as a quick idea:

```
apiVersion: capsule.clastix.io/v1
kind: Tenant
[...]

PR #871, opened 2023-10-28
feat: implement support for podOptions and additionalMetadata

gchiesa:
This PR implements the support for `additionalMetadata` enforced on Pods, similarly to the `additionalMetadata` for services.

## Rationale
In the context of soft-tenancy model, a practical use-case is represented by the need to enforce annotations / labels on Pods so that the information captured by shared cluster level services can be enriched with Tenant domain metadata (e.g. tenant-id, cost-center, etc.)

### A practical use-case 1: Scalyr/Dataset Logs
Logging agents are deployed in a cluster governed by Capsule. Multiple tenants are present. Our requirement is to be able to query logs captured from pods part of a specific tenant.
Scalyr offers the possibility to enrich the logs captured from each Pod by [leveraging annotations](https://app.scalyr.com/help/scalyr-agent-k8s#attaching).

With this functionality Pod will be automatically annotated with the information decided by cluster administrator. This will offer the possibility to categorize/query logs by tenant attributes.

### A practical use-case 2: Prometheus Metrics 
Similarly to the use case above, also prometheus metrics can be enriched with tenant level attributes that will translate in [meta labels](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#pod). 

With this functionality and using a proper Prometheus relabeling configuration the pod label/annotation can be translated in metric attributes attached to the specific pod.

## How To
Cluster administrators can configure the annotation/labels for the Pods created under a specific tenancies by using the following: 

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  owners:
    - name: alice
      kind: User
  podOptions:
    additionalMetadata:
      annotations:
        # example for scalyr logging
        log.config.scalyr.com/attributes.tenant-name: oil
      labels:
        tenant-name: oil
        tenant-cost-center: "12345"
        slack: #oil-notifications

```

## References

closes https://github.com/projectcapsule/capsule/issues/756


Makefile
```diff
@@ -330,6 +331,8 @@ e2e-install:
 		--create-namespace \
 		--set 'manager.image.pullPolicy=Never' \
 		--set 'manager.resources=null'\
+		--set "manager.image.registry=$(REGISTRY)" \
+		--set "manager.image.repository=$(REPOSITORY)" \
```
oliverbaehler, line 335: Is there a reason you added this?

oliverbaehler (changes requested):
Thanks for your awesome pr, generally it looks good to me. We will have to discuss, if it would also make sense to implement #737 along-side with this feature. WDYT @prometherion 

Please sign of all your commits.

Makefile
gchiesa: I was testing my fork with my own registry, and the `e2e` flow with helm install does not support the REGISTRY and REPOSITORY env variables when set.
If i do 
```
export REGISTRY=docker.io
export REPOSITORY=gchiesa/capsuleproject/capsule
make e2e/1.27.2
```
I was getting an error because the image was build and published but  helm release was still trying to deploy the capsule upstream image coordinates.

controllers/podlabels/abstract.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, NewNoPodMetadata(namespacedName.Name)
+	}
+
+	return tenant, nil
+}
+
+func (r *abstractPodLabelsReconciler) sync(available map[string]string, tenantSpec map[string]string) map[string]string {
+	if tenantSpec != nil {
+		if available == nil {
+			available = tenantSpec
```
prometherion, line 89: You could directly return the `available` variable here.

controllers/podlabels/pod.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type PodLabelsReconciler struct {
+	abstractPodLabelsReconciler
+
+	Log logr.Logger
+}
```
prometherion, line 14: I think we don't need the abstract here, since the Pod API is just a single one.

prometherion (approved):
Thanks @gchiesa for nailing this, as we agreed I helped in some refactoring of your code, this is the stepping stone for the v0.5.0 release!

PR #879, opened 2023-11-01
chore(security-insights): self-assessment evidence

prometherion:
According to the [CLOMonitor documentation](https://clomonitor.io/docs/topics/checks/#self-assessment):

> The details of the security self-assessment (including the evidence url) are available in the security-artifacts > self-assessment section of the [OpenSSF Security Insights](https://github.com/ossf/security-insights-spec/blob/v1.0.0/specification.md) manifest file (SECURITY-INSIGHTS.yml) that should be located at the root of the repository.


oliverbaehler (approved):
LGTM

PR #910, opened 2023-11-25
feat(tenant): add label with tenant name for each tenant

oliverbaehler:
Initial efforts for the implementation of:
https://github.com/projectcapsule/capsule-proxy/issues/328

### DIFF
--- a/controllers/tenant/metadata.go
+++ b/controllers/tenant/metadata.go
@@ -1,0 +1,23 @@
+// Copyright 2020-2023 Project Capsule Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package tenant
+
+import (
+	"context"
+
+	capsulev1beta2 "github.com/projectcapsule/capsule/api/v1beta2"
+	capsuleapi "github.com/projectcapsule/capsule/pkg/api"
+)
+
+// Sets a label on the Tenant object with it's name.
+func (r *Manager) ensureMetadata(ctx context.Context, tnt *capsulev1beta2.Tenant) (err error) {
+	// Assign Labels
+	if tnt.Labels == nil {
+		tnt.Labels = make(map[string]string)
+	}
+
+	tnt.Labels[capsuleapi.TenantNameLabel] = tnt.Name
+
+	return r.Client.Update(ctx, tnt)
+}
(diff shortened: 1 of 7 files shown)


controllers/tenant/metadata.go
```diff
@@ -0,0 +1,23 @@
...
+
+const (
+	TenantNameLabel = "kubernetes.io/metadata.name"
+)
+
+// Sets a label on the Tenant object with it's name.
+func (r *Manager) ensureMetadata(ctx context.Context, tnt *capsulev1beta2.Tenant) (err error) {
+	tnt.SetLabels(map[string]string{
+		TenantNameLabel: tnt.Name,
+	})
```
prometherion, line 20: We should keep the already existing labels, otherwise, we're gonna overwrite the user ones.

prometherion (changes requested):
I dunno why we would need a new recipe, `lint` when we already have `golint`.

controllers/tenant/metadata.go
oliverbaehler: I don't agree, using capsule specific labels seems like we are creating just a technical dept for the future. You could say the same thing here, no?
https://github.com/projectcapsule/capsule/pull/911

I would prefer using a well-known label here as well.

oliverbaehler (commented):
Tried to play around with pre-commit. But it's annoying. Removed the target.

controllers/tenant/metadata.go
prometherion: I didn't explain properly, here.

Let's say I have a Tenant:
```
NAME    STATE    NAMESPACE QUOTA   NAMESPACE COUNT   NODE SELECTOR   AGE
solar   Active                     0                                 15s
```

I'm labeling it for several reasons:
```
$: kubectl label tnt solar my-label=value
tenant.capsule.clastix.io/solar labeled

$: kubectl get tnt --show-labels
NAME    STATE    NAMESPACE QUOTA   NAMESPACE COUNT   NODE SELECTOR   AGE   LABELS
solar   Active                     0                                 49s   my-label=value
```

With the `tnt.SetLabels` provided by your code we're gonna overwrite the user labels, such as `my-label`.

oliverbaehler: @prometherion Ah right, sorry for the missunderstanding!

prometherion (approved):

PR #915, opened 2023-11-30
fix(controller): respect group quotas for number of operating system threads allocated to goroutines

oliverbaehler:
If the GOMAXPROCS environment variable is not explicitly set when starting the Go process, it defaults to the output of the runtime.NumCPU() function, which represents the number of available CPU cores (i.e., the value of P). 

**Verify** 

To verify if that's true, i added the following to main.go:

```
func MaxParallelism() int {
	maxProcs := goRuntime.GOMAXPROCS(0)
	numCPU := goRuntime.NumCPU()
	if maxProcs < numCPU {
		return maxProcs
	}
	return numCPU
}

//nolint:maintidx
func main() {
	fmt.Println("MaxParallelism: ", MaxParallelism())
```

This print out how many cores will be allocated.

1. Deployed with the main build

```
kubectl logs -f capsule-controller-manager-8c89798fc-hv5gp -n capsule-system
MaxParallelism:  4
...
```

2. Deployed with the `go.uber.org/automaxprocs` import:

```
2023/11/30 09:52:53 maxprocs: Updating GOMAXPROCS=1: using minimum allowed GOMAXPROCS
MaxParallelism:  1
``` 

In both cases the controller had the following resources allocated:

```
manager:
  resources:
    limits:
      cpu: 500m
      memory: 512Mi
    requests:
      cpu: 200m
      memory: 128Mi
```

This only takes affect, if `resources.limits.cpu` are set. If no resources are set, the amount of host CPU cores are used:

```
2023/11/30 09:50:05 maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
MaxParallelism:  4
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,6 +59,7 @@ require (
 	github.com/prometheus/common v0.44.0 // indirect
 	github.com/prometheus/procfs v0.11.1 // indirect
 	github.com/valyala/bytebufferpool v1.0.0 // indirect
+	go.uber.org/automaxprocs v1.5.3 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/exp v0.0.0-20230905200255-921286631fa9 // indirect
 	golang.org/x/net v0.17.0 // indirect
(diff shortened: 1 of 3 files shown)


prometherion (approved):
Wondering how we couldn't implement this, you the best, Oliver! 🎩

PR #935, opened 2023-12-22
docs(links): Fixed dead links

larsgerber:
Fixes the problem described in https://github.com/projectcapsule/capsule/issues/933.


prometherion (approved):
Thanks a lot, LGTM!

PR #944, opened 2024-01-12
chore(adopters): add EPAM Delivery Platform as adopter

NikolayMarusenko:


prometherion (approved):
Terrific having new Capsule public adopters, and welcome! 🫂

PR #999, opened 2024-02-28
feat(chart): add capsule-proxy 0.6.0 as optional dependency 

oliverbaehler:
Adds Capsule-Proxy as subchart. Another notable changes is, that resources are unset by default. This is the standard for any helm chart. All the other changes help to make the chart-testing work.

### DIFF
--- a/charts/capsule/Chart.lock
+++ b/charts/capsule/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: capsule-proxy
+  repository: oci://ghcr.io/projectcapsule/charts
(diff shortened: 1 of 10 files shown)


charts/capsule/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
```
maxgio92, line 1: Love it <3

maxgio92:
Thank you @oliverbaehler! 
LGTM

maxgio92 (approved):

PR #1003, opened 2024-03-05
feat: add clusterresources interception configuration

oliverbaehler:
While most of the changes happen in the capsule-proxy (see the corresponding pull request: ) the owner spec for all the tenants is extended and the capsule-proxy uses this api definition.

With this change you have additional Control of cluster Resources

```
kind: Tenant
metadata:
  name: oil
spec:
  owners:
  - kind: User
    name: alice
    clusterResources:
    - apiGroups: 
      - kyverno.io/*
      operations:
      - "List"
      resources:
      - "*"
      selector:
        matchLabels:
          consumer: tenant
    - apiGroups: 
      - admissionregistration.k8s.io/v1
      resources:
      - "mutatingwebhookconfigurations"
      selector:
        matchLabels:
          consumer: tenant
```

Note all the fields are marked as required. By default the List Operation is injected. We expect a selector to be set. This is not directly enforced with this approach, however it's the easiest way to go about this with clear documentation. Setting no selectors will have no impact on the proxy. If no requirements can be extracted, the `doesnotexist` `Exists` requirement is added. Therefor preventing from listing all the resources.


prometherion (commented):
I think we should start creating a separation of concerns between Capsule and its Proxy: since this is strictly related to the latter, shouldn't be better to move this logic there, as we did with the `ProxySetting`?

oliverbaehler:
@prometherion thanks for your review. Sure thing, but that would mean with a new stable release we would also remove the proxysettings on the ownerspec, correct?

prometherion:
I'd say so, yes, especially considering we want to act as a framework.

PR #1013, opened 2024-03-22
feat: add flag --workers

oliverbaehler:
With the workers flag you can configure the concurrent reconciles on relevant controllers. This is interesting for large cluster infrastructures


prometherion (commented):
This is interesting, although it could lead to potential issues since multiple reconciliations at the same time could generate errors due to updating of the same resource at the same time.

What I did with Kamaji was rely on a file-lock system, even tho this could be a bit more complicated: https://github.com/clastix/kamaji/pull/255

It would be perfect if we could rely on it, honestly, the license is compatible with Apache2.

WDYT @oliverbaehler?

oliverbaehler:
@prometherion Sure thing, i will implement a locking mechanism

PR #1059, opened 2024-04-30
fix(controller): ensure iteration on capsule ownerreferences

oliverbaehler:
While automating capsule for a HCI solution i noticed we have this bug. If you try to add another ownerreference you are always met with this error:

```
error: namespaces "cluster-fx9kggjwjd" could not be patched: admission webhook "namespaces.capsule.clastix.io" denied the request: Tenant.capsule.clastix.io "fx9kggjwjd" not found
```

`fx9kggjwjd` is representative for the new ownerreference name added. This fix checks if the ownerreference is actually from a tenant, if not it's skipped.

### DIFF
--- a/pkg/webhook/namespace/utils.go
+++ b/pkg/webhook/namespace/utils.go
@@ -1,0 +1,27 @@
+// Copyright 2020-2023 Project Capsule Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package namespace
+
+import (
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	capsulev1beta2 "github.com/projectcapsule/capsule/api/v1beta2"
+)
+
+const (
+	ObjectReferenceTenantKind = "Tenant"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/webhook/namespace/utils.go
```diff
@@ -0,0 +1,28 @@
...
+func isTenantOwnerReference(or metav1.OwnerReference) bool {
+	parts := strings.Split(or.APIVersion, "/")
+	if len(parts) != 2 {
+		return false
+	}
+
+	group := parts[0]
+	version := parts[1]
+
+	return group == capsulev1beta2.GroupVersion.Group && version == capsulev1beta2.GroupVersion.Version
```
prometherion, line 27: Could be this a problem with previous Capsule versions? I'm referring to the `Version`.

oliverbaehler: Yeah, we can also only check the Group

oliverbaehler: Changed the check, verify against group and kind. Disregarding version

Issue #1063, opened 2024-05-02
feat(distribution): deprecate installer


oliverbaehler:
We don't want to distribute capsule via installer.yaml files. Therefor we can deprecate all actions regarding to the installer yaml and add more resiliency for the chart. For most actions (e2e) (dev-setup) we already use the helm chart.

Issue #998, opened 2024-02-28
feat: change soft references to clastix.io


oliverbaehler:
# Describe the feature

We can start migrating to `projectcapsule.dev` from `capsule.clastix.io`. This is just preparation for a future v1 release. Places i can think of where these references can be changed without big changes:

  * `Webhook-Names`
  * `Default userGroups`
  * `E2E tests`
  * `Documentation`

# What would the new user story look like?

How would the new interaction with Capsule look like? E.g.

1. What are the prerequisites for this?
2. Tenant owner creates a new _Namespace_
3. This is going to be attached to the _Tenant_
4. All the magic happens in the background

Feel free to add a diagram if that helps explain things.

# Expected behavior
A clear and concise description of what you expect to happen.

prometherion:
Marking this as a breaking change since users must be aware of the renaming of the default user group from `capsule.clastix.io` to `projectcapsule.dev`.

PR #1065, opened 2024-05-03
feat(helm): introduce new installation and CRD lifecycle

oliverbaehler:
With this PR we are addressing the distribution of our CRDs and are trying to reconcile the Helm Chart as the main way of installing Capsule. With this change we can keep our overheads little and focus on one way of distribution. While implementing the CRDs i also did some refactoring of certain helm values. 

Implements:
-  #1063
- #998


.goreleaser.yml
```diff
@@ -29,6 +29,13 @@ builds:
           -X main.BuildTime={{ .Date }}
           -X main.GitRepo={{ .ProjectName }}
 release:
+  extra_files:
+    - glob: ./path/to/file.txt
```
maxgio92, line 33: Are these just sample files?

hack/argocd.yaml
maxgio92, line 1: Is this just a sample? Because I don't think we want to ship these manifests. WDYT?

MaxFedotov:
Hi @oliverbaehler,

what do you think about the method with a job that giantswarm uses to update crds -  https://github.com/giantswarm/kube-prometheus-stack-app/tree/a8315a8/helm/prometheus-operator-app/templates/crd-install. Seems like that will help to achieve what you want, but will be completely backward-compatible.

prometherion:
+1 for @MaxFedotov approach, it's the same approach used by HAProxy Technologies for their Kubernetes Ingress Controller.

prometherion (approved):

maxgio92 (approved):
LGTM

PR #1095, opened 2024-05-27
feat: multiple ownerreferences

oliverbaehler:


prometherion (approved):
LGTM, I'd say the controller reference here is more correct compared to the owner reference since we're not generating the object upon the Tenant creation.

PR #1148, opened 2024-08-01
fix(helm): don't render CapsuleConfiguration if crds.exclusive is set

### LINKED ISSUE #1147: Default capsuleConfiguration is rendered in helm chart when setting crds.exclusive=true
# Bug description

When the user sets `crds.exclusive=true` for capsule helm chart, default `CapsuleConfiguration` is rendered.

# How to reproduce

```
helm template charts/capsule --set crds.exclusive=true 
```

# Expected behavior

According to the comment in values.yaml
```
  # -- Only install the CRDs, no other primitives
  exclusive: false
```

only CRDs should be rendered

MaxFedotov:
closes #1147


prometherion (approved):
Definitely a bug, LGTM but please, address the CI complaints before getting it merged.

PR #1169, opened 2024-08-20
fix(controller): use typed structs for watches

oliverbaehler:
Adds support for latests Kubernetes Releases 1.30 and 1.31, removes support for older releases.


prometherion:
@oliverbaehler I faced a similar issue about generated manifests, we just need to bump up the `controller-gen` version to v0.16.1

PR #1176, opened 2024-08-26
fix(webhook) Fix regexp

coffeecupjapan:
Thank you for reading my PR.

## What?

According to kubernete's DNS validation code [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L7281-L7284), valid way to validate dns name should be 

```go
	dnsLabelFormat                 = `[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?`
	dnsSubdomainFormat             = fmt.Sprintf(`^%s(?:\.%s)*$`, dnsLabelFormat, dnsLabelFormat)
	validWindowsUserDomainDNSRegex = regexp.MustCompile(dnsSubdomainFormat)
```

I wonder we should remove `_` / should add ending `[a-zA-Z0-9]` and more small fixes or not, but in this case I just add heading `[a-zA-Z0-9]` to prevent 

``` go
`.somedns.com`
`-somedns.com`
`_somedns.com`
```

to be matched.

Or we can just imitate original kubernete code may be.

Any suggestions are welcomed. Thanks.

### DIFF
--- a/pkg/webhook/pod/containerregistry_registry.go
+++ b/pkg/webhook/pod/containerregistry_registry.go
@@ -51,7 +51,7 @@ func (r registry) Tag() string {
 
 func NewRegistry(value string) Registry {
 	reg := make(registry)
-	r := regexp.MustCompile(`((?P<registry>[a-zA-Z0-9-._]+(:\d+)?)\/)?(?P<repository>.*\/)?(?P<image>[a-zA-Z0-9-._]+:(?P<tag>[a-zA-Z0-9-._]+))?`)
+	r := regexp.MustCompile(`((?P<registry>[a-zA-Z0-9][a-zA-Z0-9-._]+(:\d+)?)\/)?(?P<repository>.*\/)?(?P<image>[a-zA-Z0-9-._]+:(?P<tag>[a-zA-Z0-9-._]+))?`)
 	match := r.FindStringSubmatch(value)
 
 	for i, name := range r.SubexpNames() {


prometherion:
Hey @coffeecupjapan, thanks for your contribution, I really appreciate it!

In regard to regex, we don't have proper unit tests and your suggested regex could be potentially broken, as shown [here](https://regex101.com/r/DVI3ci/1) for `.somedns.com/clastix/capsule:v1.0.0`.

This is definitely an edge case since it's broken: it would be great if we could have a regex that compiles only with a valid fully qualified container reference.

coffeecupjapan:
@prometherion 
Thank you for your reply.

I want to confirm where this regex is broken.
For ***.somedns.com/clastix/capsule:v1.0.0*** case `registry` group is not matched and [code here](https://github.com/projectcapsule/capsule/blob/59e5ace956de28a69cff718ad0c0144b89b969e9/pkg/webhook/pod/containerregistry.go#L88) just throws error, and I think this behavior is something expected.

Is registry for ***.somedns.com/clastix/capsule:v1.0.0*** should be valid?

Maybe I am wrong somehow, so I want to hear what I am wrong for this regexp.
Sorry for my lack of consideration.

Thank you.

prometherion:
No need to be sorry, @coffeecupjapan :)

Any enhancement or improvement here is highly welcome! As I said, I'm not the regex expert, if you think we could enhance it I'll be happy to test it and get it merged, otherwise, let's close this one to avoid having stale PR.

Just a friendly reminder, we also have a [#capsule](https://kubernetes.slack.com/archives/C03GETTJQRL) channel for direct interactions, happy to see you there!

coffeecupjapan:
@prometherion 
Thank you!
Let us discuss this topic at Slack!

PR #1213, opened 2024-10-07
feat(api): add conditions field to status

oliverbaehler:
Currently it does not do much, However this field allows transmitting conditions onto the tenant from other addons and automations.


prometherion:
This is good considering the eventually consistent nature of Kubernetes, however, I'm not very open to getting a field that is not currently used.

oliverbaehler:
Sure, will add status codes for capsule

PR #1231, opened 2024-10-23
feat(api): add tenant funcs to retrieve subjects based on clusterrole bindings

oliverbaehler:
… bindings

Found these functions to be very helpful when automating against capsule. For explenation:

With these functions we can get the subjects of a tenant (no difference in subject mapping if owner or in additionalRoleBindings) and their matching clusterRoles. This helps to translate or use tenant mapped roles to translate into other rbac implementations.

### DIFF
--- a/api/v1beta2/tenant_func.go
+++ b/api/v1beta2/tenant_func.go
@@ -4,9 +4,13 @@
 package v1beta2
 
 import (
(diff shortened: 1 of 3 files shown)


prometherion (approved):
Really helpful, LGTM!

PR #1261, opened 2024-11-24
feat: use renovate instead of dependabot

oliverbaehler:
I would like to switch to renovate, the given configuration already adheres to our semantic commits, in addition it will do the following:

- Group all github-action updates into one pr
- Group all minor and patch updates into one pr

For the rest there will be dedicated prs, specifically for this repo:

- helm dependency (capsule-proxy)
- go major bumps (i think commit digests are also considered majors)

Renovate offers much more in respect of simplifying the prs and the ecosystem which is supported. Therefor im proposing the use of renovate and the removal of dependabot.


prometherion:
I don't have a strong opinion on this, let's ping other maintainer @bsctl @MaxFedotov @maxgio92

renovate.json
```diff
@@ -0,0 +1,27 @@
...
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended", ":dependencyDashboard"],
+  "baseBranches": ["main"],
+  "prHourlyLimit": 0,
+  "prConcurrentLimit": 0,
+  "branchConcurrentLimit": 0,
+  "mode": "full",
+  "commitMessageLowerCase": "auto",
+  "semanticCommits": "enabled",
+  "semanticCommitType": "feat",
```
maxgio92, line 11: I would prefer using `chore` as we don't know in advance what's going to be introduced with a bump. WDYT?

renovate.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended", ":dependencyDashboard"],
```
maxgio92, line 3: Do we have a plan for using the `dependencyDashboard`? This is interesting ❤️

maxgio92:
I've just left some comments. I'm ok for it :)
+1

renovate.json
oliverbaehler: We had also feat(deps) with dependabot.

renovate.json
oliverbaehler: I can pin it

maxgio92 (approved):

PR #1325, opened 2025-01-19
ci: generate seccomp profile within pipeline

alegrey91:
This PR introduces the ability of generating [**seccomp**](https://kubernetes.io/docs/tutorials/security/seccomp/) profiles within the pipeline.
This will be helpful in case some user wants to secure his capsule installation by leveraging the power of **seccomp**.

To do so, `capsule` container image is built replacing the base image with the one provided by [`harpoon`](https://github.com/alegrey91/harpoon). This will allow the execution of `capsule` through `harpoon`.
Then, set up a Kubernetes cluster using `kind` with a shared folder that is used to retrieve the metadata generated by `harpoon` during the e2e tests execution.
Capsule image is consequently installed on the cluster with a customized values file from `helm`, and the e2e could start.
Once finished, the results are merged with the tracing of the unit tests (handled by `harpoon`).
This generates the `seccomp` profile tailored for `capsule`, which is going to be uploaded with the other artifacts in the upcoming releases.


Makefile
```diff
@@ -222,6 +236,12 @@ KO_VERSION = v0.14.1
 ko:
 	$(call go-install-tool,$(KO),github.com/google/ko@$(KO_VERSION))
 
+HARPOON         := /usr/sbin/harpoon
+HARPOON_VERSION := v0.9.4
+harpoon: ## Download harpoon locally if necessary.
+	@curl -s https://raw.githubusercontent.com/alegrey91/harpoon/main/install | \
+		sudo bash -s -- --install-version $(HARPOON_VERSION)
```
oliverbaehler, line 243: Not a big fan of this, we should try to install binaries into the project's bin/ folder, i am assuming this is going to install harpoon at system level. Since it,s go, can't we make a target like the other binary dependencies:

```
HARPOON = $(shell pwd)/bin/harpoon
HARPOON_VERSION = v0.9.4
harpoon:
	$(call go-install-tool,$(HARPOON),github.com/alegrey91/harpoon@$(HARPOON_VERSION))
```

charts/capsule/values-tracing.yaml
oliverbaehler, line 1: move this file to the `charts/capsule/ci` folder and call it `tracing-values.yaml`. Only specify the values necessary. This way the chart is also dry-installed with your newly added values.

charts/capsule/templates/deployment.yaml
```diff
@@ -85,12 +96,19 @@ spec:
...
+          {{- toYaml .Values.manager.volumeMounts | nindent 12 }}
+          {{- end }}
           resources:
             {{- toYaml .Values.manager.resources | nindent 12 }}
           securityContext:
+            {{- if .Values.manager.securityContext }}
+            {{- toYaml .Values.manager.securityContext | nindent 12 }}
+            {{- else }}
             {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- end }}
```
oliverbaehler, line 112: Just use `securityContext` ? Or what's the reason here

charts/capsule/templates/deployment.yaml
```diff
@@ -37,7 +37,15 @@ spec:
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
       {{- end }}
+      {{- if .Values.manager.hostPID }}
+      hostPID: {{ .Values.manager.hostPID }}
+      {{- else }}
+      hostPID: false
+      {{- end }}
+      {{- if eq .Values.priorityClassName "" }}
+      {{- else }}
```
oliverbaehler, line 46: ?

oliverbaehler (changes requested):
Hey @alegrey91 great tool, hope we can get you gain some traction as example. I have requested changes regarding the chart

Makefile
alegrey91: Hi @oliverbaehler, unfortunately I don't think we can currently use `go install` with `harpoon`.
The project has a complex toolchain that requires a separated compilation of the ebpf object.
About the installation directory, I think this is fine, I can set the location through the `install` script.

charts/capsule/templates/deployment.yaml
alegrey91: `.Values.securityContext` is used by `kubelet` image.
`.Values.manager.securityContext` is used only by the capsule image when built for tracing.
When we don't set `.Values.manager.securityContext`, then `.Values.securityContext` come into play for both the images.
Does it make sense for you?

charts/capsule/templates/deployment.yaml
alegrey91: Maybe I might remove it. Was probably a typo.

charts/capsule/templates/deployment.yaml
oliverbaehler: Which `kubelet` image? We just have the manager that we are deploying..

ARe you talking about this here:

```
trace-install: trace-load-image
	helm upgrade \
	    --dependency-update \
		--debug \
		--install \
		--namespace capsule-system \
		--create-namespace \
		--set 'manager.image.pullPolicy=Never' \
		--set 'manager.image.registry=localhost' \
		--set 'manager.image.repository=capsule-tracing' \
		--set 'manager.resources=null'\
		--set "manager.image.tag=latest" \
		--set "manager.image.securityContext.privileged=true" \
		--set 'manager.livenessProbe.failureThreshold=10' \
		--set 'manager.readinessProbe.failureThreshold=10' \
		--values charts/capsule/values-tracing.yaml \
		capsule \
		./charts/capsule
```

This specifcally `manager.image.securityContext.privileged=true` (This path is wrong btw)?  So the rest of the securitycontext is too restrictive? Otherwise make sure it points to the same values in the Makefile as in the chart values

oliverbaehler (changes requested):

charts/capsule/templates/deployment.yaml
alegrey91: Sorry, I meant the `kubectl` image which is mentioned in the `values.yaml`.
The point is that `.Values.securityContext` is shared between the capsule `Deployment` and the kubectl `Job`, so in order to be able to overwrite only the capsule `Deployment` I created a dedicated item for it.

alegrey91:
Hello @oliverbaehler, I'm having a problem with the Test charts pipeline.
It's failing due to my values file `ci/tracing-values.yaml` because this specific configuration needs **kind** to be configured with an `extraMount` (see `hack/kind-cluster.yml`). Without this cluster config, the chart is unable to mount the volumes specified within the values file.
What can we do to fix it?

charts/capsule/templates/deployment.yaml
oliverbaehler: Jobs take sec from `.Vales.global.jobs.kubectl.securityContext`, they are no longer shared

alegrey91: Sorry, actually the reason why I created `.Values.manager.securityContext` was that when I overwrite the default values with `tracing-values.yaml`, this result in the following setting:
```
          securityContext:
            capabilities:
              add:
              - SYS_ADMIN
              - NET_ADMIN
              - PERFOM
              drop:
              - ALL
            privileged: true
            readOnlyRootFilesystem: true
```
This is a combination of both (`values.yaml`, `ci/tracing-values.yaml`) that doesn't make sense.
(`privilege: true` and `allowPrivilegeEscalation: false` cannot be used at the same time).
For this reason I created `.Values.manager.securityContext`, so I could specify a different `securityContext` in particular cases where I need it.

alegrey91:
> Hello @oliverbaehler, I'm having a problem with the Test charts pipeline. It's failing due to my values file `ci/tracing-values.yaml` because this specific configuration needs **kind** to be configured with an `extraMount` (see `hack/kind-cluster.yml`). Without this cluster config, the chart is unable to mount the volumes specified within the values file. What can we do to fix it?

I'm thinking about a possible solution for that. Since the `ci/tracing-values.yaml` is not a "production" values file, we could skip its test with `ct` and create a dedicated workflow to generate `seccomp` profiles for each PR. This way we will check its validity every time a new PR is coming. Integrate it with `ct` is quite hard because there are several pre-requirements:
* build the custom capsule image with harpoon within the container
* create a kind cluster with a custom configuration to share folders with the container

oliverbaehler:
@alegrey91 can you grant me access to your fork?

alegrey91:
> @alegrey91 can you grant me access to your fork?

You should have access now.

oliverbaehler:
@alegrey91 I can't invest more time into this, sorry. I heave reconciled your targets and simplified the approach. It should now work better.

alegrey91:
@oliverbaehler any reasons why the other workflows are not starting anymore?

oliverbaehler:
@alegrey91 conflicts?

alegrey91:
@oliverbaehler ready for review.
thanks for the fixes :)

.github/workflows/seccomp.yaml
```diff
@@ -0,0 +1,66 @@
...
+    paths:
+      - '.github/workflows/e2e.yml'
+      - 'api/**'
+      - 'controllers/**'
+      - 'pkg/**'
+      - 'e2e/*'
+      - 'Dockerfile'
+      - 'go.*'
+      - 'main.go'
+      - 'Makefile'
```
oliverbaehler, line 28: Here only on pull-requests, could you also adjust it in the e2e.yaml that it does also only execute on pr. This should save us some resources and time :D

oliverbaehler (changes requested):
Final changes, sowy :3

.github/workflows/seccomp.yaml
alegrey91: Sure, makes sense

alegrey91:
@oliverbaehler can you please re-run the CI?

oliverbaehler (approved):

PR #1339, opened 2025-01-31
feat: Add additionalMetadataList Support for Conditional Metadata Assignment

Deofex:
**Summary**
This PR introduces support for additionalMetadataList, allowing cluster admins to define conditional labels and annotations for namespaces within a tenant. This enhancement builds upon the existing additionalMetadata functionality, providing more granular control through namespaceSelector expressions.

**Changes Introduced**

- Added support for additionalMetadataList in namespaceOptions, enabling selective metadata assignment based on namespace labels.
- Maintains existing additionalMetadata behavior, ensuring all namespaces receive global metadata if specified.
- Introduced namespaceSelector-based rules, allowing metadata to be applied only to namespaces that match specific conditions.
- Updated documentation to clarify the new functionality
- If multiple additionalMetadataList entries match a namespace, all applicable labels and annotations are assigned.

**Example Configuration**
Defining both global and conditional metadata in a tenant:

```yaml
kind: Tenant
metadata:
  name: oil
spec:
  owners:
    - name: alice
       kind: User
  namespaceOptions:
    additionalMetadata:
      labels:
        capsule.clastix.io/still-working: "true"
    additionalMetadataList:
      - labels:
          capsule.clastix.io/backup: "true"
      - namespaceSelector:
          matchExpressions:
            - key: capsule.clastix.io/low_security_profile
              operator: NotIn
              values: ["true"]
         labels:
           pod-security.kubernetes.io/enforce: baseline
      - namespaceSelector:
          matchExpressions:
            - key: capsule.clastix.io/low_security_profile
              operator: In
              values: ["true"]
         labels:
           pod-security.kubernetes.io/enforce: privileged
```

**Impact & Compatibility**
Fully backward compatible: Existing additionalMetadata configurations will continue to work as before.
New functionality is opt-in, meaning it does not affect existing tenants unless additionalMetadataList is explicitly configured.


oliverbaehler:
We'll have to do some discussions around this. I am currently refactoring to a generic CEL based approach which would also solve this case. But thanks for the commit for now, I will get back to it

oliverbaehler:
@peterbosalliandercom yes i get your use-case and it does make sense. However we want to move away from such specific implementations and refactor into more generic solutions, since they are technical dept. Such that we could eg. deprecated the additionalMetadata for pod, services etc. as well. A generic approach to having patches for any kind of manifests on tenant basis with given selector options. That's what i would aim for. 

I will have to create a poc on the weekend and see how difficult that is to implement. Based on these findings we will either merge this as is and keep the prospect, that it will be deprecated in the v1 api bump or i will port the same functionality to that generic feature.

Make sense?

prometherion:
> Is there any way we can speedup this feature request?

Generally speaking, we're always happy to receive newer contributions: we have limited resources to allocate to the project, especially considering the efforts in reviewing code from non-maintainers.

Remember that Capsule is backed by two vendors (CLASTIX and PeakScale) which offer features prioritisations as well as other professional services, and addons.

> I am currently refactoring to a generic CEL based approach which would also solve this case.

We should go toward CEL and drop support of custom handlers where it is possible: @maxgio92 started a [PoC](https://github.com/maxgio92/capsule-policy-vap) with VAP/CEL while working at CLASTIX.

oliverbaehler:
Sorry i haven't had time yet. I am happy to accept this change in the current spec. @Deofex could you merge and resolve the conflicts on the crd?

controllers/tenant/namespaces.go
```diff
@@ -73,6 +73,25 @@ func (r *Manager) syncNamespaceMetadata(ctx context.Context, namespace string, t
...
+						return err
+					}
+
+					if !namespaceSelectedBySelector {
+						continue
+					}
+
+					for k, v := range additionalMetadata.Labels {
+						labels[k] = v
+					}
```
oliverbaehler, line 91: Add annotations as well

```suggestion
					for k, v := range additionalMetadata.Annotations{
						annotations[k] = v
					}
			}
```

controllers/tenant/namespaces.go
```diff
@@ -73,6 +73,25 @@ func (r *Manager) syncNamespaceMetadata(ctx context.Context, namespace string, t
...
+				}
+			}
+
```
oliverbaehler, line 94: Initialize the labels empty on L59

```
labels:= make(map[string]string)
```

Add default labels here, to ensure they were not overwriten by spec:
```suggestion
			labels["kubernetes.io/metadata.name"] = namespace
			labels[capsuleLabel] =  tnt.GetName()
```

charts/capsule/crds/capsule.clastix.io_globaltenantresources.yaml
oliverbaehler, line 1: this should not be touched with your change

pkg/api/additional_metadata.go
```diff
@@ -3,9 +3,12 @@
...
+
 // +kubebuilder:object:generate=true
 
 type AdditionalMetadataSpec struct {
-	Labels      map[string]string `json:"labels,omitempty"`
-	Annotations map[string]string `json:"annotations,omitempty"`
+	NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+	Labels            map[string]string     `json:"labels,omitempty"`
+	Annotations       map[string]string     `json:"annotations,omitempty"`
 }
```
oliverbaehler, line 14: Create a new spec for this, otherwise the other apis also have a nemspaceSelector field out of context.

```suggestion
	Labels            map[string]string     `json:"labels,omitempty"`
	Annotations       map[string]string     `json:"annotations,omitempty"`
}

type AdditionalMetadataSelectorSpec struct
	NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
	Labels            map[string]string     `json:"labels,omitempty"`
	Annotations       map[string]string     `json:"annotations,omitempty"`
}

```

api/v1beta2/namespace_options.go
```diff
@@ -13,6 +13,8 @@ type NamespaceOptions struct {
 	Quota *int32 `json:"quota,omitempty"`
 	// Specifies additional labels and annotations the Capsule operator places on any Namespace resource in the Tenant. Optional.
 	AdditionalMetadata *api.AdditionalMetadataSpec `json:"additionalMetadata,omitempty"`
+	// Specifies additional labels and annotations the Capsule operator places on any Namespace resource in the Tenant via a list. Optional.
+	AdditionalMetadataList []api.AdditionalMetadataSpec `json:"additionalMetadataList,omitempty"`
```
oliverbaehler, line 17:
```suggestion
	AdditionalMetadataList []api.AdditionalMetadataSelectorSpec `json:"additionalMetadataList,omitempty"`
```

oliverbaehler (changes requested):

api/v1beta2/namespace_options.go
Deofex: Changed

pkg/api/additional_metadata.go
Deofex: Added the spec

controllers/tenant/namespaces.go
Deofex: Made the change, also added a test to check this

controllers/tenant/namespaces.go
Deofex: Added

Svarrogh1337:
@Deofex workflows approved. Lets see a green e2e :)

Deofex:
There's quite a bit of red there, unfortunately.

Classic “it works on my machine” 🙃:

```
Ran 115 of 129 Specs in 644.815 seconds  
SUCCESS! -- 115 Passed | 0 Failed | 0 Pending | 14 Skipped
```

I'll take a look, but if anyone has an idea of what’s causing the failures, I’d appreciate the input — I’m not seeing anything in my code that would explain them.

I also noticed the linter is failing on a Cyclomatic Complexity check. I’ll see if I can split up the `syncNamespaceMetadata` method.

Is the `check commit` a required check? If so, I’ll need to adjust my commit messages accordingly.

Deofex:
I refactored the ```syncNamespaceMetadata``` method. 
I adjusted my commit messages.
The e2e test succeeds multiple times on my laptop. I can't find the cause why it failed in the pipeline. 

Can someone review the refactored method and try the 2e2 test again in the pipeline?

oliverbaehler (approved):
Thank you very much, i'm going to extend the docs later

PR #1345, opened 2025-02-05
chore(repo): add pre-commit

oliverbaehler:
I have noticed a lot of users having troubles with commit messaging etc. For first time contributors they have to wait until we approve the runs again. To prevent this, i have added some pre-commit hooks which keep the semantics of files clean, check commit messages, generate helm related stuff, and lint the code. This helps users to detect errors already locally without any further action.

I have also added a schema (schema generator target) for our helm chart.


prometherion (approved):
Lovely, thanks a lot Oliver!

PR #1379, opened 2025-02-16
feat: namespace metadata sync on creation #1378

### LINKED ISSUE #1378: Sync namespace metadata on creation.
We need to sync namespace metadata (labels, annotation and especially nodeSelector) on namespace creation.
For now, namespace nodeSelector reconciled by controller loop and pod can start before nodeSelector will be set on ns (in case of huge amount of namespaces).

Let's add some extra logic to namespace webhook.

*Have prepared PR for this already.

CharlieR-o-o-t:
Closes #1378

Sync metadata on namespace creation

### DIFF
--- a/pkg/webhook/namespace/mutation/cordoning.go
+++ b/pkg/webhook/namespace/mutation/cordoning.go
@@ -68,7 +68,9 @@ func (h *cordoningLabelHandler) syncNamespaceCordonLabel(ctx context.Context, c 
 
 	if label, ok := ns.Labels[ln]; ok {
 		if err = c.Get(ctx, types.NamespacedName{Name: label}, tnt); err != nil {
-			admission.Errored(http.StatusInternalServerError, err)
+			response := admission.Errored(http.StatusInternalServerError, err)
+
+			return &response
 		}
 	}
 
--- a/pkg/webhook/namespace/mutation/cordoning.go
+++ b/pkg/webhook/namespace/mutation/cordoning.go
@@ -6,10 +6,11 @@ package mutation
 import (
 	"context"
 	"encoding/json"
+	"net/http"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
-	"net/http"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -48,13 +49,16 @@ func (h *cordoningLabelHandler) OnUpdate(c client.Client, decoder admission.Deco
 		if err := decoder.Decode(req, ns); err != nil {
 			return utils.ErroredResponse(err)
 		}
+
 		response := h.syncNamespaceCordonLabel(ctx, c, req, ns)
+
 		return response
 	}
 }
 
 func (h *cordoningLabelHandler) syncNamespaceCordonLabel(ctx context.Context, c client.Client, req admission.Request, ns *corev1.Namespace) *admission.Response {
 	tnt := &capsulev1beta2.Tenant{}
+
 	ln, err := capsuleutils.GetTypeLabel(tnt)
 	if err != nil {
 		response := admission.Errored(http.StatusInternalServerError, err)
--- a/pkg/webhook/namespace/mutation/metadata.go
+++ b/pkg/webhook/namespace/mutation/metadata.go
@@ -1,0 +1,81 @@
+// Copyright 2020-2023 Project Capsule Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package mutation
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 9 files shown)


CharlieR-o-o-t:
Hello @oliverbaehler 
Apologies for the delay, was on vacation.

At first look, I thought this change would be a quick 5-minute adventure. But soon I have realized that all mutation changes need to be handled in a single place. I have also added e2e for both webhook and controller namespace reconcile.

Additionally, I encountered issues with the Harpoon image:
ghcr.io/alegrey91/harpoon:latest is a broken build—there is no execute bit set on the binary, causing the container to crash during trace-e2e. As a workaround, I switched Harpoon back to Docker Hub.

Please take a look when you'll have time. 
pipeline should be ok this time.

CharlieR-o-o-t:
Have no failed test in trace-e2e on my ubuntu workstation. 
Doesn't look like related to my change. Will figure out.

Svarrogh1337:
> Additionally, I encountered issues with the Harpoon image: ghcr.io/alegrey91/harpoon:latest is a broken build—there is no execute bit set on the binary, causing the container to crash during trace-e2e. As a workaround, I switched Harpoon back to Docker Hub.
> 
> Please take a look when you'll have time. pipeline should be ok this time.

You can `git cherry-pick f06c38f280bcf524d6b58b172d4e86800c25f138` for fix. It's also already been merged into main, so you can take it from there too.

CharlieR-o-o-t:
@Svarrogh1337  @oliverbaehler , apologies, just found time for this.
Latest harpoon produces deadlock, I can reproduce it only in github CI. 
I just tried to remove harpoon from entrypoint for test purposes and everything is okay. 

On my ubuntu workstation everything is ok even with harpoon:
Screenshot_20250428_011221

pkg/webhook/namespace/patch.go
oliverbaehler, line 1: Why does the logic have to be in one handler?

The `patch.go` is a validating handler. Therefor it should not be combined with your change. You have done the right thing with registering a new route to generalize the mutation of namespace. However i would rename the current ownerreference folder into namespace-mutating or whatever and then keep the logic for patching in a dedicated handler (as is) and add a new handler, which just handles the nodeSelector.

CharlieR-o-o-t: Got your point about `patch.go`.

There is a huge logic to detect namespace ownership in onCreate, I just wanted to do it Once in one place.

CharlieR-o-o-t: @oliverbaehler , could you take a look at new PR. Is it what you wanted?

oliverbaehler:
Yeah there was just a feagure merged which adds the subpackages for namespace mutations and validations. You may add your logic now as dedicated handler in the namespcae mutation package. Should be much simpler now

CharlieR-o-o-t:
@oliverbaehler , I have made changes considering last mutation feature, could you take a look?

pkg/webhook/namespace/mutation/cordoning.go
oliverbaehler, line 1: You may delete this one, as yours achieves the same thing

pkg/webhook/namespace/mutation/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+		if tenant == nil {
+			response := admission.Denied("Unable to assign namespace to tenant.")
+
+			return &response
+		}
+
+		// sync namespace metadata
+		capsuletenant.SyncNamespaceMetadata(tenant, ns)
+
+		return nil
```
oliverbaehler, line 53: You still need to patch the admission, this does not do anything yet:

```suggestion
		// sync namespace metadata
		capsuletenant.SyncNamespaceMetadata(tenant, ns)
	
	        marshaled, err := json.Marshal(ns)
	        if err != nil {
		  response := admission.Errored(http.StatusInternalServerError, err)
		  return &response
	        }
	        
	        response := admission.PatchResponseFromRaw(req.Object.Raw, marshaled)
	        return &response
```

oliverbaehler (changes requested):

oliverbaehler:
I'm. really sorry @CharlieR-o-o-t , i just merged the other feature which also causes some conflicts since it changes some of the metadata code. I would just put everything into your sync function so it's also considered at admission create, sahouldn't be too big of a change

pkg/webhook/namespace/mutation/metadata.go
CharlieR-o-o-t: > You still need to patch the admission, this does not do anything yet:

yes, sorry, night coding on vacation is evil), patch was lost in changes. fixed

CharlieR-o-o-t:
@oliverbaehler , 
Everything mentioned were fixed. Thank you for review.

I have also added additional commit on cordon webhook (26dbdf08b2ab). 
Internal error was not returned.

oliverbaehler (approved):
Thank you very much for your efforts! 🚀

oliverbaehler:
Mh? Not sure what you are looking for, Cordoning does not even run `OnCreate`? 

`InCapsuleGroups` tells the webhook, that the user submitting the admission (or rather one of their groups) must be mentioned in the `capsuleCOnfiguration`. Otherwise the admission is just allowed.  This is not to interfere with non tenant operations.

CharlieR-o-o-t:
I mean that no matter how many handlers added to InCapsuleGroups, executed only first. 
This breaks the logic with mutation.

PR #1414, opened 2025-04-08
fix(ci): remove harpoon from workflow

oliverbaehler:


prometherion:
Just making aware of this @alegrey91, hope he can promptly fix Harpoon to keep it on the CI

Svarrogh1337:
@alegrey91, I think I got the the bottom of it - LMKWYT:
```Go
func decodeInstruction(instHex []byte) ([]uint64, error) {
	var offsets []uint64
	s := bytes.NewBufferString("")
	for i := 0; i < len(instHex); {
		inst, err := x86asm.Decode(instHex[i:], 64)
		fmt.Printf("%04X\t%s\n", i, inst.String())
		if err != nil {
			fmt.Printf("Error on %d iteration: %v\n", i, err)
		}
		s.WriteString(fmt.Sprintf("%04X\t%s", i, inst.String()))
		s.WriteString("\n")
		fmt.Println(inst.Mode)
		if err != nil {
			return nil, err
		}
		if inst.Op == x86asm.RET {
			offsets = append(offsets, uint64(i))
		}
		i += inst.Len
	}
	return offsets, nil
}
```
In this function we are attempting to Decoding in 64-bit mode, but some of the instructions are incompatible with that mode. For example below I am getting AAS which is 32-bit mode instruction - [documentation](https://www.felixcloutier.com/x86/aas). From there the offset is broken.
| 64-bit mode  | 32-bit mode |
| ------------- | ------------- |
| 0000    NOP  | 0000    NOP  |
| 0001    OR EAX, [RAX-0x7] | 0001    OR EAX, [RAX-0x7] |
|0004    ICEBP | 0004    ICEBP |
|0005    AND [RSI], -0x2f| 0005    AND [RSI], -0x2f |
|0008    Op(0)| 0008    AAS|

alegrey91:
@Svarrogh1337 good catch!
Yes, this might definitely be the cause of the problem.
We could try of managing both 64 and 32 bit within the function.
Do you want to take care of this?

Svarrogh1337:
@alegrey91 lets catchup in slack

oliverbaehler:
E2E is no longer working reliable. We are going through with this change.

alegrey91:
> E2E is no longer working reliable. We are going through with this change.

Sure, I'm sorry we are not being able to fix the problem. Locally I didn't have these problems. Maybe this is due to the limited resources on the runner.

Svarrogh1337:
> > E2E is no longer working reliable. We are going through with this change.
> 
> Sure, I'm sorry we are not being able to fix the problem. Locally I didn't have these problems. Maybe this is due to the limited resources on the runner.

If we bump the timeout reasonably and combine it with https://github.com/projectcapsule/capsule/pull/1428, we should be good, but I will try it over next few days. Running e2e locally works fine for me, so it must be related to the runner.

PR #1415, opened 2025-04-09
fix(ci): make harpoon binary executable

Svarrogh1337:
fix(ci): make harpoon binary executable


oliverbaehler:
@Svarrogh1337 Were you ableto build the image locally?

Svarrogh1337:
@oliverbaehler yes, I had to find device with amd64 arch and test the fix there,but it worked. For the moment harpoon does not support arm/arm64 due to the usage of  "golang.org/x/arch/x86/x86asm" for decoding of the instructions.

PR #1424, opened 2025-04-25
feat(tenant): improve cordoning selection

Svarrogh1337:
resolves https://github.com/projectcapsule/capsule/issues/1421


controllers/tenant/namespaces.go
```diff
@@ -73,6 +73,12 @@ func (r *Manager) syncNamespaceMetadata(ctx context.Context, namespace string, t
 				}
 			}
 
+			if tnt.Spec.Cordoned {
+				labels["projectcapsule.dev/cordoned"] = "true"
+			} else {
+				labels["projectcapsule.dev/cordoned"] = "false"
+			}
+
```
oliverbaehler, line 81: We don't want a label, when it's not cordoned. Also with only this, you will cause clashes with gitops engines, as they will also try to patch this away. You should additionally verify at admission, if this label must be set, if patch.

you can add a new handler here:
https://github.com/projectcapsule/capsule/tree/main/pkg/webhook/ownerreference

pkg/webhook/namespace/patching.go
```diff
@@ -0,0 +1,345 @@
...
+		if err = c.Get(ctx, types.NamespacedName{Name: label}, tnt); err != nil {
+			admission.Errored(http.StatusInternalServerError, err)
+		}
+	}
+
+	if !tnt.Spec.Cordoned {
+		delete(ns.Labels, "projectcapsule.dev/cordoned")
+	} else {
+		ns.Labels["projectcapsule.dev/cordoned"] = "true"
+	}
```
oliverbaehler, line 328: only handle the true case at admission. deletion should be done by the controller.

pkg/webhook/namespace/patching.go
oliverbaehler, line 1: create a dedicated file instead of using the `patching.go`. This keeps a better maintainability of the code

controllers/tenant/namespaces.go
```diff
@@ -73,6 +73,10 @@ func (r *Manager) syncNamespaceMetadata(ctx context.Context, namespace string, t
 				}
 			}
 
+			if tnt.Spec.Cordoned {
+				labels["projectcapsule.dev/cordoned"] = "true"
+			}
+
 			if tnt.Spec.NodeSelector != nil {
```
oliverbaehler, line 80: Use here the if from the admission

oliverbaehler (changes requested):

pkg/webhook/namespace/patching.go
Svarrogh1337: I've merged [ownerreference/patching](https://github.com/Svarrogh1337/capsule/blob/9ce677d9e4ef1a5be9ee485ecbab7faa76b6602e/pkg/webhook/ownerreference/patching.go#L1-L319) into the new namespaces/patching.go. I will revert it to them being separated.

Svarrogh1337:
@oliverbaehler all comments are now resolved.

charts/capsule/templates/mutatingwebhookconfiguration.yaml
```diff
@@ -81,28 +81,29 @@ webhooks:
...
+    {{- include "capsule.webhooks.service" (dict "path" "/namespace-patch" "ctx" $) | nindent 4 }}
   failurePolicy: {{ .failurePolicy }}
   matchPolicy: Equivalent
-  name: owner.namespace.projectcapsule.dev
-  namespaceSelector: {}
-  objectSelector: {}
+  name: namespace-patching.tenants.projectcapsule.dev
+  namespaceSelector:
+  {{- toYaml .namespaceSelector | nindent 4}}
+  objectSelector: { }
```
oliverbaehler, line 95:
```suggestion
  {{- with .namespaceSelector }} 
  namespaceSelector:
    {{- toYaml . |  nindent 4 }}
  {{- end }}
  {{- with .objectSelector }} 
  objectSelector:
    {{- toYaml . |  nindent 4 }}
  {{- end }}
```

controllers/tenant/namespaces.go
```diff
@@ -73,6 +73,12 @@ func (r *Manager) syncNamespaceMetadata(ctx context.Context, namespace string, t
 				}
 			}
 
+			if tnt.Spec.Cordoned {
+				ns.Labels["projectcapsule.dev/cordoned"] = "true"
```
oliverbaehler, line 77: offload the label to somewhere more central, so we have an overview what metadata we are using. I would suggest:

https://github.com/projectcapsule/capsule/blob/main/pkg/utils/tenant_labels.go

but in this case it can just be a constant

pkg/webhook/namespace/patching.go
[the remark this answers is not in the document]
oliverbaehler: Just to clarify, you have done it correctly, but i would prefer to have a bit more context by the folder naming and having the same routes grouped by directory

main.go
```diff
@@ -6,6 +6,7 @@ package main
 import (
 	goflag "flag"
 	"fmt"
+	"github.com/projectcapsule/capsule/pkg/webhook/ownerreference"
```
oliverbaehler, line 9: Project imports are always part of the first block. Import ordering:

1. Std-Libs
2. 3-Party
3. Imports from this project

charts/capsule/values.yaml
```diff
@@ -267,12 +267,22 @@ webhooks:
...
     namespaceOwnerReference:
       failurePolicy: Fail
+    customresources:
+      failurePolicy: Fail
+      namespaceSelector:
+        matchExpressions:
+          - key: capsule.clastix.io/tenant
+            operator: Exists
+    namespacePatching:
+      failurePolicy: Fail
```
oliverbaehler, line 277:
```suggestion
    customresources:
      failurePolicy: Fail
      namespaceSelector:
        matchExpressions:
          - key: capsule.clastix.io/tenant
            operator: Exists
      objectSelector: {} 
    namespacePatching:
      failurePolicy: Fail
      namespaceSelector: {}
      objectSelector: {}
      
```

charts/capsule/templates/validatingwebhookconfiguration.yaml
```diff
@@ -274,4 +274,31 @@ webhooks:
...
+    - v1
+    - v1beta1
+  clientConfig:
+    {{- include "capsule.webhooks.service" (dict "path" "/customresources" "ctx" $) | nindent 4 }}
+  failurePolicy: {{ .failurePolicy }}
+  matchPolicy: Equivalent
+  name: customresources.tenant.projectcapsule.dev
+  namespaceSelector:
+  {{- toYaml .namespaceSelector | nindent 4}}
+  objectSelector: { }
```
oliverbaehler, line 288:
```suggestion
  {{- with .namespaceSelector }} 
  namespaceSelector:
    {{- toYaml . |  nindent 4 }}
  {{- end }}
  {{- with .objectSelector }} 
  objectSelector:
    {{- toYaml . |  nindent 4 }}
  {{- end }}
```

Svarrogh1337:
@oliverbaehler all has been resolved.
e2e tests were also added to cover the label.
https://github.com/projectcapsule/capsule/blob/f51e47b6571e768c25886ae081a564c8f0e3d355/e2e/tenant_cordoning_test.go#L109-L120
https://github.com/projectcapsule/capsule/blob/f51e47b6571e768c25886ae081a564c8f0e3d355/e2e/tenant_cordoning_test.go#L72-L83

oliverbaehler (changes requested):

Svarrogh1337:
@oliverbaehler these are resolved as well.

pkg/webhook/namespace/mutation/cordoning.go
```diff
@@ -0,0 +1,77 @@
...
+		if err = c.Get(ctx, types.NamespacedName{Name: label}, tnt); err != nil {
+			admission.Errored(http.StatusInternalServerError, err)
+		}
+	}
+
+	if tnt.Spec.Cordoned {
+		ns.Labels[capsuleutils.CordonedLabel] = "true"
+	}
+
+	return nil
```
oliverbaehler, line 76: This actually does not do anything, these changes must be unmarshaled and then patched to the response

```suggestion
	if !tnt.Spec.Cordoned {
	  return nil
	}
	
	labels := ns.GetLabels()
	if _, ok := labels[capsuleutils.CordonedLabel]; !ok {
	  return nil
	}
	
        c, err := json.Marshal(ns)
	if err != nil {
		return admission.Errored(http.StatusInternalServerError, err)
	}
	
	return admission.PatchResponseFromRaw(req.Object.Raw, c)
```

Implement correctly with IDE support :D

oliverbaehler (approved):
Thanks for your patience 😄

PR #1429, opened 2025-04-29
fix(ci): switch to large runner for e2e tests

Svarrogh1337:


Svarrogh1337:
@prometherion @oliverbaehler, that's the middle ground between runner size and timeout. 
Once we have more e2e pipelines runs I can analyse the data and make further adjustments. 
Let me know if you have any questions.

oliverbaehler (approved):
Thanks for the efforts ❤️

PR #1463, opened 2025-05-18
feat(tenant): support gateway/class

Svarrogh1337:
feat(tenant): support gateway/class


api/v1beta2/gateway_options.go
```diff
@@ -0,0 +1,12 @@
...
+
+package v1beta2
+
+import (
+	"github.com/projectcapsule/capsule/pkg/api"
+)
+
+type GatewayOptions struct {
+	AllowedClasses *api.DefaultAllowedListSpec `json:"allowedClasses,omitempty"`
+}
```
oliverbaehler, line 12: We should no longer use this, the exactMatch and Regex should no longer be used. Instead, in `pkg/api/allowed_list.go` add the following new struct:

```go
// +kubebuilder:object:generate=true

type SelectionListWithDefaultSpec struct {
	SelectorAllowedListSpec `json:",inline"`
	Default                 string `json:"default,omitempty"`
}

func (in *SelectionListWithDefaultSpec) MatchDefault(value string) bool {
	return in.Default == value
}

// +kubebuilder:object:generate=true

type SelectionListWithSpec struct {
	metav1.LabelSelector `json:",inline"`
}

func (in *SelectionListWithSpec) MatchSelectByName(obj client.Object) bool {
	if obj != nil {
		return in.Match(obj.GetName()) || in.SelectorMatch(obj)
	}

	return false
}

func (in *SelectionListWithSpec) SelectorMatch(obj client.Object) bool {
	if obj != nil {
		selector, err := metav1.LabelSelectorAsSelector(&in.LabelSelector)
		if err != nil {
			return false
		}

		return selector.Matches(labels.Set(obj.GetLabels()))
	}

	return false
}
```

This will be the future API

pkg/webhook/gateway/validate_class.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil
+	case allowed.Match(gatewayClass.Name) || selector:
+		return nil
+	default:
+		recorder.Eventf(tnt, corev1.EventTypeWarning, "ForbiddenGatewaClass", "Gateway %s/%s GatewayClass %s is forbidden for the current Tenant", req.Namespace, req.Name, &gatewayClass)
+
+		response := admission.Denied(NewGatewayClassForbidden(gatewayObj.Name, *allowed).Error())
+
+		return &response
+	}
```
oliverbaehler, line 114:
```suggestion
	case allowed.MatchDefault(gatewayClass.Name):
		return nil
	case selector:
		return nil
	default:
		recorder.Eventf(tnt, corev1.EventTypeWarning, "ForbiddenGatewaClass", "Gateway %s/%s GatewayClass %s is forbidden for the current Tenant", req.Namespace, req.Name, &gatewayClass)

		response := admission.Denied(NewGatewayClassForbidden(gatewayObj.Name, *allowed).Error())

		return &response
	}
```

oliverbaehler (changes requested):

api/v1beta2/gateway_options.go
Svarrogh1337: @oliverbaehler if I follow the logic, shouldn't we embed _SelectionListWithSpec_ in _SelectionListWithDefaultSpec_ instead of _SelectorAllowedListSpec_ or I am missing something?

oliverbaehler: yeah sorry

Svarrogh1337:
@oliverbaehler I did switchover and tested few scenarios and it looks good. If you have something else in mind let me know, else I'd be moving to e2e tests definitions.

Makefile
```diff
@@ -224,8 +224,12 @@ golint: golangci-lint
 e2e: ginkgo
 	$(MAKE) e2e-build && $(MAKE) e2e-exec && $(MAKE) e2e-destroy
 
+e2e-install-deps:
+	@$(KUBECTL) apply --force-conflicts --server-side=true -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.0/standard-install.yaml
```
oliverbaehler, line 228: This should auto update api gateway

```suggestion
API_GW         := none
API_GW_VERSION := v1.3.0
API_GW_LOOKUP  := kubernetes-sigs/gateway-api/
e2e-install-deps:
	@$(KUBECTL) apply --force-conflicts --server-side=true -f https://github.com/$(API_GW_LOOKUP)/releases/download/$(API_GW_VERSION)/standard-install.yaml
```

Svarrogh1337:
@oliverbaehler I am done here. We are now not denying on validation non-existent GatewayClassess instead of mutating them to default as discussed and it's validated by the e2e tests. We can apply the same logic for Ingressess as well.

oliverbaehler (approved):

PR #1477, opened 2025-05-22
chore: promote hristo hristov as maintainer

oliverbaehler:
He has shown that he has what it takes to be a maintainer. I greatly appreciate his help and we share the same mindset. I am more than happy to have him on board. 

**Note**: No reaction within a week from other maintainers is valued as an approval


prometherion (approved):
Hristo is doing a great job, I couldn't agree more on this!

PR #1517, opened 2025-06-25
feat(tenant): expose additional metrics

### LINKED ISSUE #1516: Add additional metrics for Tenant
# Describe the feature

We want to add additional metrics for Tenant:

1. Amount of namespaces (size)
2. For each associate namespace (via label)
3. Tenant state (active/cordoned)

# Expected behavior
The listed Metrics are exposed and can be used for monitoring.

Svarrogh1337:
resolves #1516


Svarrogh1337:
@oliverbaehler I did refactor the approach. LMK WYT

controllers/tenant/metrics.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Exposing Status Metrics for tenant.
+func (r *Manager) syncStatusMetrics(tenant *capsulev1beta2.Tenant) {
+	var cordoned float64 = 0
+	// Reset all metrics
+	r.Metrics.DeleteTenantStatusMetrics(tenant.GetName())
+	// Expose namespace-tenant relationship
+	for _, ns := range tenant.Status.Namespaces {
+		r.Metrics.TenantNamespaceRelationshipGauge.WithLabelValues(tenant.GetName(), ns).Set(1)
+	}
```
oliverbaehler, line 15: Can you think of a smoother implementation than always deleting all of the exemplers? Even if so, `DeletePartialMatch` is very resource intensive. We should avoid doing that on every loop for scalability. However you might be able to preserve the status namespaces before reconciling and diffing between the namespaces and remove those, which are no longer in the status

controllers/tenant/manager.go
```diff
@@ -91,6 +91,10 @@ func (r Manager) Reconcile(ctx context.Context, request ctrl.Request) (result ct
 
 		return
 	}
+	// Ensuring Status metrics are exposed
+	r.Log.Info("Ensuring all status metrics are exposed")
+	r.syncStatusMetrics(instance)
```
oliverbaehler, line 96: Why is this before the loop? This wont be an accurate representation?

oliverbaehler (changes requested):

oliverbaehler (changes requested):

PR #1546, opened 2025-07-16
feat: add toggle for securitycontext and podsecuritycontext

sandert-k8s:
Added toggles for enabling/disabling securityContext and PodSecurityContext. If you have other security measurements in place (for example Openshift SCC's), the default SecurityContexts are not needed. In the past, we provided an empty string for the securityContext, which always worked. But suddenly, ArgoCD does not accept this anymore. So, adding a toggle for this.

(This is the same change as added in the [sops-operator](https://github.com/peak-scale/sops-operator/pull/90) and [capsule-proxy](https://github.com/projectcapsule/capsule-proxy/pull/796) )

### DIFF
--- a/charts/capsule/Chart.yaml
+++ b/charts/capsule/Chart.yaml
@@ -41,4 +41,4 @@ annotations:
       url: https://projectcapsule.dev/
   artifacthub.io/changes: |
     - kind: added
-      description: oci chart reference
+      description: added toggles for podSecurityContexts and securityContexts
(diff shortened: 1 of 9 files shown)


prometherion:
Just saw projectcapsule/capsule-proxy#796, which is pretty similar: I think it's safe to get merged, it also maintains a backward compatibility in terms of previous behaviour.

sandert-k8s:
> Just saw [projectcapsule/capsule-proxy#796](https://github.com/projectcapsule/capsule-proxy/pull/796), which is pretty similar: I think it's safe to get merged, it also maintains a backward compatibility in terms of previous behaviour.

Yes, I've linked the other PR's in the opening post; they are all the same. I think this one isn't merged yet because there was a pipeline failing, which I fixed after that.

PR #1612, opened 2025-08-25
fix: ensure empty arrays rendered for userNames and ignoreUserWithGroups

cz-bs:
Without this fix, Helm emits null when userNames or ignoreUserWithGroups are empty or unset. This breaks CRD validation (e.g. in Argo CD) because the CapsuleConfiguration schema requires arrays if the keys are present. This patch ensures valid YAML by defaulting to empty arrays.

### DIFF
--- a/charts/capsule/templates/configuration-default.yaml
+++ b/charts/capsule/templates/configuration-default.yaml
@@ -20,13 +20,21 @@ spec:
   {{- range .Values.manager.options.capsuleUserGroups }}
     - {{ . }}
   {{- end }}
+  {{- if .Values.manager.options.userNames }}
   userNames:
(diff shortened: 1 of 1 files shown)


oliverbaehler:
Will be implemented by #1614, are you using serverside apply with client side diff?

cz-bs:
> Will be implemented by #1614, are you using serverside apply with client side diff?

exactly:
- ssa on this applicationset
- client side diff (argocd's default)

oliverbaehler:
@cz-bs you may try upgrading to 0.10.8

cz-bs:
> @cz-bs you may try upgrading to 0.10.8

Fixed, thanks!

PR #1702, opened 2025-10-14
fix(chart): add argocd annotations for crd lifecycle job

zvlb:
This PR adds an ArgoCD hook annotation to ensure CRDs are created/updated before other resources during ArgoCD deployments.

**Problem**:
When deploying Capsule via ArgoCD, CRDs are applied simultaneously with other resources, causing installation failures when custom resources are processed before CRDs are ready.

**Solution**:
The PreSync hook ensures CRDs are properly installed and available before ArgoCD proceeds with syncing the remaining resources, enabling successful Capsule deployment in ArgoCD-managed clusters.

More informations about ArgoCD sync phases [here](https://argo-cd.readthedocs.io/en/latest/user-guide/sync-waves/)

### DIFF
--- a/charts/capsule/templates/crd-lifecycle/_helpers.tpl
+++ b/charts/capsule/templates/crd-lifecycle/_helpers.tpl
@@ -4,6 +4,7 @@
 
 {{- define "capsule.crds.annotations" -}}
 "helm.sh/hook": "pre-install,pre-upgrade"
+"argocd.argoproj.io/hook": "PreSync"
 {{- end }}
 
 {{- define "capsule.crds.component" -}}


zvlb:
[Same](https://github.com/projectcapsule/capsule-proxy/pull/905) for capsule proxy

sandert-k8s:
Hi, 

I'm unsure if this will do/fix anything, since ArgoCD already maps these helm hooks to argocd hooks, see https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks . 
When using the crd install job (which is the recommended way of installation), you should [disable the installation of crd's through argocd](https://argo-cd.readthedocs.io/en/latest/user-guide/helm/#helm-skip-crds).
And in the helm chart, enable the crd job (by default enabled)
```
crds:
  install: true
```

Does this solve the issue? What kind of issues do you encounter when setting/not setting these hook annotations?

zvlb:
I'm not sure if the annotation
"[helm.sh/hook](https://helm.sh/hook)": "pre-install,pre-upgrade"
is being processed correctly by ArgoCD.
Perhaps it needs to have only one value specified?

When I use
```
crds:
  install: true
```
My ArgoCD workflows start running before the job that installs the CRDs is executed, and everything fails.

sandert-k8s:
Did you disable the installation of the CRD's in the ArgoCD Application spec? https://argo-cd.readthedocs.io/en/latest/user-guide/helm/#helm-skip-crds

sandert-k8s:
I think I've spotted the problem. ArgoCD does a dry sync of all resources, and when the installation of the CRD's is set to false in the ArgoCD application spec, it fails to Sync on the initial sync becauses it tries a dryrun on the `CapsuleConfiguration` which isn't yet possible. 

Could you set
```
customAnnotations:
  argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
```
in the values of Capsule? I think the initial installation should succeed then.

oliverbaehler:
#1710 Adds more specific values to add `argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true` to only the configuration object. We will add that to the doc. @zvlb Your feature already exists, you can use the values for `.Values.global.jobs.kubectl.annotations`

oliverbaehler:
https://github.com/projectcapsule/capsule/blob/b531a8628d9efef5a7f3b558b14a771ef47b571a/charts/capsule/templates/crd-lifecycle/job.yaml#L13

PR #1749, opened 2025-11-20
feat(config): administrators get delete privileges for tenant namespaces

oliverbaehler:


internal/controllers/utils/name_matching.go
```diff
@@ -4,11 +4,36 @@
...
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		oldObj, ok1 := e.ObjectOld.(*capsulev1beta2.CapsuleConfiguration)
+		newObj, ok2 := e.ObjectNew.(*capsulev1beta2.CapsuleConfiguration)
+		if !ok1 || !ok2 {
+			return false
+		}
+
+		return !reflect.DeepEqual(oldObj.Spec, newObj.Spec)
+		// or with k8s helpers:
+		// return !equality.Semantic.DeepEqual(oldObj.Spec, newObj.Spec)
```
prometherion: We should avoid at our best the usage of the `reflection` package since it's expensive.

PR #1767, opened 2025-12-02
feat(config): add combined users property as successor for usergroups

oliverbaehler:
Also adds proper deprecation warnings etc.


Svarrogh1337 (approved):
E2E failure is false positive. Else lgtm

PR #1912, opened 2026-05-08
preview: add pvc selector support

oliverbaehler:


hack/generated/kubeconfig.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
```
prometherion, line 1: Should be this committed @oliverbaehler ?

PR #1927, opened 2026-05-29
fix: reuse webhookport from values

oliverbaehler:


Svarrogh1337 (approved):
LGTM

bakito:
It must be assured, that the port on the dynamic-webhook matches the pod and service ports.

oliverbaehler:
@bakito agrre with the changes?

bakito:
> @bakito agrre with the changes?

no. The default port should also be considered here:
- https://github.com/projectcapsule/capsule/blob/main/charts/capsule/templates/_helpers.tpl#L164
- https://github.com/projectcapsule/capsule/blob/main/charts/capsule/templates/_helpers.tpl#L181

otherwise generated dynamic webhooks are pointing to the wrong service port

bakito:
```
{{/*
Capsule Webhook service (Called with $.Path)

*/}}
{{- define "capsule.webhooks.service" -}}
  {{- include "capsule.webhooks.cabundle" $.ctx | nindent 0 }}
  {{- if $.ctx.Values.webhooks.service.url }}
url: {{ printf "%s/%s" (trimSuffix "/" $.ctx.Values.webhooks.service.url ) (trimPrefix "/" (required "Path is required for the function" $.path)) }}
  {{- else }}
service:
  name: {{ default (printf "%s-webhook-service" (include "capsule.fullname" $.ctx)) $.ctx.Values.webhooks.service.name }}
  namespace: {{ default $.ctx.Release.Namespace $.ctx.Values.webhooks.service.namespace }}
  port: {{ default $.ctx.Values.manager.webhookPort $.ctx.Values.webhooks.service.port }}
  path: {{ required "Path is required for the function" $.path }}
  {{- end }}
{{- end }}


{{/*
Capsule Webhook service (Without Path)
*/}}
{{- define "capsule.webhooks.serviceConfig" -}}
  {{- include "capsule.webhooks.cabundle" $ | nindent 0 }}
  {{- if $.Values.webhooks.service.url }}
url: {{ trimSuffix "/" $.Values.webhooks.service.url }}
  {{- else }}
service:
  name: {{ default (printf "%s-webhook-service" (include "capsule.fullname" $)) $.Values.webhooks.service.name }}
  namespace: {{ default $.Release.Namespace $.Values.webhooks.service.namespace }}
  port: {{ default $.Values.manager.webhookPort $.Values.webhooks.service.port }}
  {{- end }}
{{- end }}
```

oliverbaehler:
@bakito didnt push. .. D:

charts/capsule/templates/_helpers.tpl
bakito, line 164: default need to be changed here too

bakito:
LGTM

PR #1928, opened 2026-05-29
fix: correct helm values schema for webservcie ports

bakito:
Correct the helm schema for webhook ports


bakito:
helm-docs did not update correctly yet, don't know how to tell the tool how to change the type or the null value.

oliverbaehler (approved):

PR #1949, opened 2026-06-04
feat(deps): bump golang 1.26.4 

oliverbaehler:


prometherion (approved):
LGTM, although we should strive for toolchain and go parity.

PR #1986, opened 2026-06-28
test: disable capsule-proxy cert-manager resources for e2e

bakito:
Local e2e test can not be started since capsule-proxy chart has cert-manager resources enabled by default.

This PR disables the cert manager resources in proxy subchart and allows the test to run again locally.


Makefile
```diff
@@ -476,6 +476,7 @@ e2e-install: helm-controller-version ko-build-all dev-install-gw-api-crds
 		--create-namespace \
 		--set 'replicaCount=2'\
 		--set 'certManager.generateCertificates=false' \
+		--set 'proxy.enabled=false' \
```
oliverbaehler: Proxy should be tested in e2e. We can reusue the make target to install cert manager for the dev-setup

oliverbaehler (changes requested):

Makefile
bakito: can we disable the cert-manager resources?

PR #1997, opened 2026-07-03
feat(globalcustomquota): add support for not-equals field selector

sandert-k8s:
Add support for not-equals feat selector in the globalcustomquota's. 
For quota's, the succeeded/unkown/failed pod's shouldn't be added since they don't claim their resources in the kubelet. With this change, you can do this for example:
```
kind: GlobalCustomQuota
metadata:
  name: cpu-limits
spec:
  namespaceSelectors:
    - matchLabels:
         tenant: solar
  sources:
      - apiVersion: v1
        kind: Pod
        op: add
        path: .spec.containers[*].resources.limits.cpu
        selectors:
          - fieldSelectors:
              - .status.phase!=Succeeded
              - .status.phase!=Failed
              - .status.phase!=Unknown
 ```

This PR is just a suggestion for a solution. Feel free to close if it's not up to quality standards.

### DIFF
--- a/e2e/customquota_global_test.go
+++ b/e2e/customquota_global_test.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/ptr"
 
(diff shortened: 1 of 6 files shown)


oliverbaehler (approved):
🚀  Don't forget to add docs here https://projectcapsule.dev/docs/resource-management/customquotas/#jsonpath

PR #2023, opened 2026-07-11
fix(webhook): let administrators repair namespaces with an inconsistent tenant reference

### LINKED ISSUE #2022: Administrators cannot repair (or delete) namespaces with an inconsistent tenant reference (label without Tenant ownerReference)
## Bug description

A namespace that carries the tenant label (`capsule.clastix.io/tenant`) **without** the matching `Tenant` ownerReference can neither be **updated** nor **deleted** by anybody — including identities listed in `CapsuleConfiguration.spec.administrators`. The inconsistency can therefore never be repaired through the API.

Such half-assigned namespaces are not hypothetical: they get created whenever a namespace is admitted **while the Capsule webhooks are unavailable** (operator outage/crashloop, `failurePolicy: Ignore` window, or objects that predate the 0.13 consistency model). We hit this in production with a namespace created during a controller outage: it was invisible to quota/NetworkPolicy replication AND impossible to fix.

## How to reproduce

1. Produce an inconsistent namespace (simulates a webhook outage leftover), e.g. scale the capsule operator to 0 with `failurePolicy: Ignore`, or as an administrator strip the ownerReference from an assigned namespace.
2. Add the repairing identity to `CapsuleConfiguration.spec.administrators` (e.g. `{kind: Group, name: system:masters}`).
3. Try to repair it as that administrator:

```bash
kubectl patch ns broken-ns --type=merge -p '{"metadata":{"ownerReferences":[{"apiVersion":"capsule.clastix.io/v1beta2","kind":"Tenant","name":"my-tenant","uid":"<tenant-uid>","controller":true}]}}'
```

**Actual result** (any update, even the one that would restore consistency — and `kubectl delete` as well):

```
Error from server: admission webhook "namespaces.validating.projectcapsule.dev" denied the request: namespace has tenant label "my-tenant" but no Tenant ownerReference
```

**Expected**: administrators should be able to repair (or delete) such namespaces — that's precisely what `administrators` is for, and the mutating `guard` handler already fully bypasses admins.

## Root cause

In `internal/webhook/namespace/validation/handler.go`, `OnUpdate` calls `tenant.ResolveNamespaceTenant()` on the **old** and new objects *unconditionally* and turns any error into a denial **before** the `user.IsAdmin()` bypass can apply (`ResolveNamespaceTenant` errors on `label != "" && len(refs) == 0`). `OnDelete` has the same pattern, so an inconsistent namespace cannot even be removed. Verified on `v0.13.9` and current `main`.

Related: #1941 addressed admin updates of tenant-**less** namespaces; this issue is about *inconsistent* ones (label without ownerReference, and the other mismatch states `ResolveNamespaceTenant` rejects).

## Workaround

Temporarily exclude the affected namespace from the `namespaces` validating hook via `webhooks.hooks.namespaces.objectSelector` (chart values), patch, then revert — clumsy for something `administrators` should allow.

## Proposed fix

Treat resolution errors as non-fatal **for administrators only** in `OnUpdate`/`OnDelete`: the inconsistent object is handled as tenant-less for them, which permits the repair (restoring the ownerReference or removing the stale label) as well as deletion. `OnCreate` is left strict (the mutating assignment webhook guarantees consistency on that path) and nothing changes for non-administrators.

PR incoming with the fix and an e2e case in `e2e/config_administrators_test.go`.

## Environment

- Capsule v0.13.9 (also present on `main`)
- Kubernetes v1.35, cert-manager-managed webhook certificates

antoinemichea:
## Summary

Fixes #2022.

A namespace carrying the tenant label without the matching `Tenant` ownerReference (the typical leftover of a webhook outage) could neither be updated nor deleted by anybody: in the namespace **validation** handler, `ResolveNamespaceTenant()` errors were turned into denials *before* the `administrators` bypass could apply — locking out even the identities that `CapsuleConfiguration.spec.administrators` is meant to empower, so the inconsistency could never be repaired through the API.

## Changes

- `internal/webhook/namespace/validation/handler.go`:
  - **OnUpdate**: resolution errors on the old/new object are only fatal for non-administrators. Administrators see the inconsistent object as tenant-less, which allows the repair itself (restoring the missing ownerReference, or removing the stale label) while every non-admin check is preserved unchanged.
  - **OnDelete**: same tolerance, so administrators can also get rid of an inconsistent namespace.
  - **OnCreate is intentionally unchanged**: on that path the mutating assignment webhook guarantees consistency, and the existing e2e (`creating namespace with faulty label`) expects strictness.
- `e2e/config_administrators_test.go`: new case *"repairing namespaces with an inconsistent tenant reference as administrator"* — makes a namespace inconsistent, asserts a tenant owner is still denied, asserts an administrator can update it, repair it (ownerReference restored → namespace back in the tenant), and delete a second inconsistent one.

## Behaviour matrix (update/delete of an inconsistent namespace)

| Identity | Before | After |
|---|---|---|
| tenant owner / capsule user | denied (error) | denied (unchanged) |
| administrator | **denied (error)** | **allowed — treated as tenant-less, repair possible** |

This mirrors the mutating `guard` handler, which already fully bypasses administrators, and follows up on #1941 (admin updates of tenant-*less* namespaces).

### DIFF
--- a/e2e/config_administrators_test.go
+++ b/e2e/config_administrators_test.go
@@ -284,4 +284,139 @@ var _ = Describe("Administrators", Ordered, Label("namespace", "permissions", "a
 		})
 
 	})
+
+	It("repairing namespaces with an inconsistent tenant reference as administrator", func() {
+		ctx := context.TODO()
+
+		ns1 := NewNamespace("", map[string]string{
+			meta.TenantLabel: tnt1.GetName(),
+		})
+
+		By("creating a namespace assigned to the tenant", func() {
+			NamespaceCreation(ns1, admin, defaultTimeoutInterval).Should(Succeed())
+			NamespaceIsPartOfTenant(tnt1, ns1).Should(Succeed())
+		})
+
+		By("stripping the Tenant ownerReference as administrator (simulates the leftover of a webhook outage)", func() {
+			Eventually(func() error {
+				current := NewNamespace(ns1.Name)
+
+				if err := k8sClient.Get(ctx, types.NamespacedName{Name: ns1.Name}, current); err != nil {
+					return err
+				}
+
+				original := current.DeepCopy()
+				current.OwnerReferences = nil
+
+				return impersonationClient(admin.Name, nil).Patch(ctx, current, client.MergeFrom(original))
+			}, defaultTimeoutInterval, defaultPollInterval).Should(Succeed())
+		})
+
+		By("verifying the namespace is inconsistent (tenant label without Tenant ownerReference)", func() {
+			current := NewNamespace(ns1.Name)
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: ns1.Name}, current)).To(Succeed())
+			Expect(current.Labels[meta.TenantLabel]).To(Equal(tnt1.GetName()))
+			Expect(current.OwnerReferences).To(BeEmpty())
+		})
+
+		By("updating the inconsistent namespace as tenant owner is still denied", func() {
+			current := NewNamespace(ns1.Name)
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: ns1.Name}, current)).To(Succeed())
+
+			original := current.DeepCopy()
+			if current.Annotations == nil {
+				current.Annotations = map[string]string{}
+			}
+
+			current.Annotations["e2e.capsule.clastix.io/owner-update"] = "true"
+
+			err := impersonationClient(tnt1.Spec.Owners[0].Name, nil).Patch(ctx, current, client.MergeFrom(original))
+			Expect(err).To(HaveOccurred())
+		})
+
+		By("updating the inconsistent namespace as administrator succeeds", func() {
+			Eventually(func() error {
+				current := NewNamespace(ns1.Name)
+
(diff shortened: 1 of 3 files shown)


antoinemichea:
Thanks @copilot-pull-request-reviewer — all four points were legitimate, addressed in the latest commit:

- **Fail-open on API errors (both inline comments)**: introduced a `tenant.ErrInconsistentTenantReference` sentinel wrapping the *semantic* mismatch cases in `ResolveNamespaceTenant` (label without ownerReference and vice versa, multiple ownerReferences, unknown Tenant, label/ownerReference mismatch). `OnUpdate`/`OnDelete` now only tolerate `errors.Is`-matching errors for administrators — transient API/status failures from the Tenant lookup keep failing closed for everybody. The "unknown Tenant" case is included on purpose so admins can also repair namespaces referencing a deleted Tenant. Existing error-message consumers are unaffected (the wrapped messages keep the original substrings).
- **e2e denial assertion**: the tenant-owner denial now asserts the message contains `no Tenant ownerReference`, ruling out unrelated failures such as RBAC.
- **e2e repair assertion**: after the repair the test now verifies, directly on the namespace, that the Tenant ownerReference is restored (kind/name) and consistent with the tenant label, in addition to the Tenant status membership.

oliverbaehler:
@antoinemichea i have implemented the fix

Issue #2041, opened 2026-07-19
Admins & external controllers can't modify namespace metadata that doesn't match forbiddenLabels & forbiddenAnnotation deny rules


tom-stockwell:
# Bug description

When a Tenant has `forbiddenLabels` or `forbiddenAnnotations` rules configured under `spec.namespaceOptions`, capsule admins and non-capsule users are blocked from modifying any metadata on tenant namespaces that does not also match these rules.

Admins and external controllers / service accounts should not be beholden to the restrictions put on tenant owners to modify metadata for their own namespaces.

# How to reproduce

Create a new Tenant like so:
```yaml
apiVersion: capsule.clastix.io/v1beta2
kind: Tenant
  name: tenant-a
spec:
  owners:
    - kind: User
      name: user1
  namespaceOptions:
    forbiddenAnnotations:
      deniedRegex: .*
    forbiddenLabels:
      deniedRegex: .*
```

Have the tenant create a namespace:
```bash
kubectl --context=admin create namespace tenant-a-1
```

Try to label the namespace as the tenant owner:
```bash
$ kubectl --context=user1 label namespace tenant-a-1 any=label
Error from server (Forbidden): admission webhook "namespaces.validating.projectcapsule.dev" denied the request: namespace labels validation failed: test is forbidden for the current Tenant. Forbidden are matching the regex .*
```

Try to label the namespace as an admin:
```bash
$ kubectl --context=admin label namespace tenant-a-1 any=label
Error from server (Forbidden): admission webhook "namespaces.validating.projectcapsule.dev" denied the request: namespace labels validation failed: test is forbidden for the current Tenant. Forbidden are matching the regex .*
```

# Expected behavior

The label command should succeed for the admin user, if they are either a non-capsule user or a capsule admin.

# Additional context

- Capsule version: v0.13.9
- Helm Chart version: 0.13.9
- Kubernetes version: v1.34.8

Svarrogh1337:
Thanks for the issue and the PR. I've enabled the workflows on the PR, let's see green pipeline : )

oliverbaehler:
This is intentional, not a bug. Please elborate on your use case and why it can't be solved via additionalMetadataList

tom-stockwell:
My specific use case is that OpenShift uses the kube-controller-manager to assign security metadata and labels to namespaces. In particular the following annotations:

- `openshift.io/sa.scc.mcs`
- `openshift.io/sa.scc.uid-range`
[...]

PR #2042, opened 2026-07-19
fix: Enforce namespace metadata validation for capsule users only

tom-stockwell:
Addresses #2041 by simply restricting the `userMetadataForValidation` path to only apply when the admission user is a capsule user.

This allows for both capsule admins and non-capsule users to modify namespace metadata without being beholden to the tenant restrictions.


tom-stockwell:
I see the problem with the e2e tests. I guess that's what I get for just eyeballing them. I'll get them running locally so I can test them before pushing up the fixes.

tom-stockwell:
e2e tests are working locally, should be good to retest

oliverbaehler (approved):
LGTM

Issue #2069, opened 2026-08-06
manager.rbac.strict is a no-op; the cluster-admin binding is gated by the undocumented manager.rbac.minimal


dva08:
## Bug description

The Capsule Helm chart v0.13.7 exposes `manager.rbac.strict: true` as a configuration option intended to run the controller without `cluster-admin` privileges. However, enabling this flag does **not** remove the `cluster-admin` ClusterRoleBinding from the controller's ServiceAccount — it only adds an aggregate ClusterRole. The controller continues to operate with `cluster-admin` permissions regardless of the flag value.

## Reproduction steps

1. Install Capsule via the Helm chart with `manager.rbac.strict: true` (chart version 0.13.7, controller image v0.13.7)
2. Check the ClusterRoleBinding for the capsule controller:

```bash
kubectl get clusterrolebinding | grep capsule-manager
# capsule-manager-rolebinding   ClusterRole/cluster-admin   ...
3. The binding is still present and references ClusterRole/cluster-admin — strict mode has no effect on it.
Expected behavior
When manager.rbac.strict: true is set, the Helm chart should either:
- Remove the cluster-admin ClusterRoleBinding entirely, OR
- Replace it with a scoped ClusterRole that grants only the permissions Capsule actually needs
Actual behavior
The capsule-manager-rolebinding ClusterRoleBinding persists and still references ClusterRole/cluster-admin, regardless of the manager.rbac.strict flag value.
Evidence
ClusterRoleBinding (confirmed on live cluster):
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: capsule-manager-rolebinding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: capsule
  namespace: capsule-system
Helm chart values.yaml (capsule-0.13.7):
rbac:
  strict: false  # <-- When set to true, expected to remove cluster-admin binding
Controller image: ghcr.io/projectcapsule/capsule:v0.13.7
All Capsule ClusterRoleBindings:
capsule-manager-rolebinding                                     ClusterRole/cluster-admin
capsule-namespace-provisioner                                   ClusterRole/capsule-namespace-provisioner
capsule-pre-delete                                              ClusterRole/capsule-pre-delete
capsule:capsule:aggregate                                       ClusterRole/capsule:capsule:aggregate
[...]

PR #2088, opened 2026-08-19
feat(chart): bind default owner clusterroles in strict rbac mode

Svarrogh1337:
resolves second part of #2069 
feat(chart): bind default owner ClusterRoles in strict RBAC mode

### DIFF
--- a/charts/capsule/templates/rbac.yaml
+++ b/charts/capsule/templates/rbac.yaml
@@ -273,6 +273,7 @@ rules:
   - "get"
   - "list"
   - "watch"
+  - "create"
   - "patch"
   - "update"
   - "delete"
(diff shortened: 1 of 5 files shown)


charts/capsule/templates/rbac.yaml
```diff
@@ -433,6 +434,14 @@ rules:
   - update
   - patch
   - delete
+- apiGroups: ["rbac.authorization.k8s.io"]
+  resources: ["clusterroles"]
+  resourceNames:
+    - admin
+    - {{ $.Values.manager.options.rbac.provisioner}}
+    - {{ $.Values.manager.options.rbac.deleter}}
```
oliverbaehler: This should probably also be variable via values, eg under `manager.rbac.bindableClusterRoles: []`

oliverbaehler (changes requested):

charts/capsule/templates/rbac.yaml
Svarrogh1337: @oliverbaehler done in the latest commit.

oliverbaehler (approved):

PR #2124, opened 2026-09-06
sec: reintroduce namespace subresources 

oliverbaehler:


bakito:
@oliverbaehler this one should be obsolete now

PR #2129, opened 2026-09-09
test: add make tasks to improve developer experience

bakito:
- introduce make task: select-laptop-host-ip to help handling of LAPTOP_HOST_IP (requires [gum](https://github.com/charmbracelet/gum))
- add missing but documented `dev-cluster` task
- update capsule version in playground
- correct helm release namespace in playground


playground/installation/capsule/release.flux.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   chart:
     spec:
       chart: capsule
-      version: "0.13.11"
+      version: "0.14.6"
```
bakito: What about using 'latest' here?

PR #2165, opened 2026-09-22
fix(api): match mixed-case allowed list entries

tikket1:
`AllowedListSpec.ExactMatch` still carries the sort/search mismatch that was fixed in `ForbiddenListSpec.ExactMatch` for GHSA-gjw4-3v3v-rqxg (0.13.7).

`pkg/api/allowed_list.go` sorts the slice with a **case-insensitive** comparator and then binary searches it with `sort.SearchStrings`, which is only correct on a slice ordered **byte-ascending**. A list mixing capitalised and lowercase entries can be ordered so that the search walks the wrong half and misses an entry that is present.

```go
sort.SliceStable(in.Exact, func(i, j int) bool {
    return strings.ToLower(in.Exact[i]) < strings.ToLower(in.Exact[j])
})

i := sort.SearchStrings(in.Exact, value)

ok = i < len(in.Exact) && in.Exact[i] == value
```

The match itself is case-sensitive (`in.Exact[i] == value`), so the `ToLower` affects only the ordering, never the result. It buys nothing and breaks the precondition of the search.

### Reproduction

```go
allowed := []string{"nginx", "NGINX", "traefik"}

ExactMatch(allowed, "nginx")   // false  <-- present in the list
ExactMatch(allowed, "NGINX")   // false  <-- present in the list
ExactMatch(allowed, "traefik") // true
```

Case-insensitive ordering puts `NGINX` before `nginx`; `sort.SearchStrings` assumes the opposite and lands on the wrong index for both.

### Impact

This is the opposite direction to the advisory, which is why it is an ordinary bug report and not a security report. On a forbidden list a missed match let a denied key through. On an allow list a missed match **denies a permitted value**.

At `internal/webhook/ingress/validate_class.go` the polarity is `case allowed.Match(x): return nil` with `default:` falling through to forbidden, so a tenant whose IngressClass is explicitly allowed can be rejected. The same matcher backs the gateway, DRA, storage-class, priority-class and container-registry allow-lists.

Second, smaller issue in the same function: `sort.SliceStable` reorders `in.Exact` **in place** on every call. That slice belongs to a cached `Tenant` / `CapsuleConfiguration` object shared across concurrent admission requests, so the sort races with other goroutines reading the same backing array. Switching to `slices.Contains` removes the mutation entirely.

### Fix

Mirrors `ForbiddenListSpec.ExactMatch` as fixed in 0.13.7, and matches `containsExact` in `pkg/api/runtime/expression.go`.

`sort` and `strings` become unused in the file and are dropped; `regexp` is still needed by `RegexMatch`.

### Test

Adds a mixed-case case to the existing `TestAllowedListSpec_ExactMatch` table. It fails on `main` (two assertions) and passes with the change. The `False` entries also pin that matching stays case-sensitive.

### Notes on scope

- **No behaviour is lost.** There is no input where the old code returns `true` and `slices.Contains` returns `false`: the old code returning `true` requires `in.Exact[i] == value`, which already implies membership. I checked this over 200k randomised mixed-case lists, 0 divergences in that direction. Nothing outside the configured list can match now that did not match before.
- **Not a performance regression.** The old path sorted the slice on *every* call, so it was `O(n log n)` per lookup against `O(n)` for a linear scan over lists that are typically a handful of entries.
- **Nothing depends on the ordering side effect.** Every other reader of `.Exact` in the tree iterates it, `strings.Join`s it, or builds a `sets.New` from it.
- The other two sort-then-`SearchStrings` sites, `HostnamesList.IsStringInList` and `pkg/users.userGroupList.Find`, sort byte-ascending and are correct. This is the only mismatched one. Happy to leave them alone or fold in a follow-up if you would rather they used `slices.Contains` too.
- `go build ./...`, `go vet`, and `go test ./pkg/... ./internal/...` (3331 tests, 89 packages) pass. `gofmt` clean.

Let me know if you would prefer this as an issue instead, or want the fix shaped differently.


oliverbaehler (approved):
Thanks for your contribution 🚀

